PR #1, opened 2019-09-20
Fix local build

VictorVelarde:
Some small tweaks I had to do to avoid build errors, when starting fresh in my local repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ It uses [Lerna](https://lerna.js.org) to manage all the packages and inter-depen
 Quick steps to check that build works correctly
 
 ```sh
+ nvm use
  npm install
  lerna bootstrap
  npm run build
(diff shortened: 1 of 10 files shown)


VictorVelarde:
Without the explicit declaration of modules, my build failed consistently... 🤷‍♂

README.md
```diff
@@ -8,3 +8,11 @@ This is a monorepo intended to contain all small-medium sized JS libraries that
 - A Metapackage that exposes all
 
 It uses [Lerna](https://lerna.js.org) to manage all the packages and inter-dependencies.
+
+Quick steps to check that build works correctly
+
+```sh
+ npm install
```
rjimenezda:
```suggestion
 nvm use
 npm install
```

packages/sql/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@carto/toolkit-sql';
```
rjimenezda: These declar should not be necessary, could you try removing them? And the references to index.d.ts

VictorVelarde: I have removed them, and it compiles. But it complains on npm run test

Issue #3, opened 2019-09-23
Manage Typescript declaration files automagically


VictorVelarde:
Right now, no **.d.ts files** are created automatically from source. As a result, some Type info between packages is not correctly detected.

eg

```
import { Credentials } from '@carto/toolkit-auth';

class Maps {
  private _credentials: any; // TODO Credentials class type is not detected
}

```

TODO: Review tsc compiling & rollup config, so Types are automatically included in the library (and / or externally published?)

PR #4, opened 2019-09-23
Add type generation for current toolkit packages

jesusbotella:
I've enabled type module generation in TSC config for all packages.

Definitions are set to be generated in `dist/types` folder, so they won't be committed to the repository but they will be generated whenever TSC is run.


VictorVelarde:
Closes https://github.com/CartoDB/toolkit/issues/3

PR #6, opened 2019-09-24
Add basic OAuth functionality

rjimenezda:
It even has tests for some of it :joy:


packages/auth/__tests__/utils.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      'dataservices:fluxcapaciting',
+      'datasets:w:wadus',
+      'datasets:metadata',
+      'schemas:c'
+    ];
+
+    const unknown = unknownScopes(scopes);
+
+    expect(unknown).toHaveLength(2);
+    expect(unknown).toContain('invent');
```
VictorVelarde: 😄 best scope, I want to have it

packages/auth/__tests__/UserInfo.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    const mockJson = Promise.resolve({
+      wadus: true,
+      username: 'roman-carto'
+    });
+
+    const mockFetch = Promise.resolve({
+      json: () => mockJson
+    });
+
+    (global as any).fetch = jest.fn().mockImplementation(() => mockFetch);
```
VictorVelarde: nice mocking

packages/auth/index.html
```diff
@@ -0,0 +1,41 @@
...
+    loginBtn.addEventListener('click', async () => {
+      const data = await oauth.login();
+      console.log('Got data from login promise:', data);
+    });
+
+    oauth.on('tokenUpdated', (d) => {
+      console.log('Token updated', oauth.credentials);
+    });
+
+    oauth.on('error', () => {
```
VictorVelarde: I guess event includes an object with the error cause, right?

packages/auth/rollup.config.ts
```diff
@@ -9,15 +9,14 @@ const { modules, umd } = require('../../rollup-config-generator');
 const pkg = require('./package.json');
 
 export default [
-  umd('auth', 'src/index.ts', [
+  umd('CartoToolkitAuth', 'src/index.ts', pkg, [
```
VictorVelarde: Why do we need to add the pkg here?

rjimenezda: I've changed the config generator, it uses the package to extract the name of the module / browser / main files.

packages/auth/src/CartoProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  get logout() {
+    return `${this.config.url}/logout`;
+  }
+
+  get userInfo() {
+    return this._userInfo;
+  }
+
+  public $validate(options: CartoValidation): void {
```
VictorVelarde: I can see the override here, but... Do we really need to keep the dollar prefix here? Isn't it a bit strange for the final user?

packages/auth/index.html
rjimenezda: I just checked, only in one case, I should change it so it always comes

packages/auth/src/CartoProvider.ts
rjimenezda: You should not call that directly imho. The only reason why it's public is because it is public on salte-auth :man_shrugging:

packages/auth/src/CartoProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Returns Expiration date for current token, or -1 if already expired
+   *
+   * @readonly
+   * @type {number}
+   * @memberof CartoProvider
+   */
+  get expiresIn(): number {
+    if (!this.accessToken || this.accessToken.expired) {
+      return -1;
```
VictorVelarde: Let's add a small beautiful const of sth. like ALREADY_EXPIRED?

packages/auth/src/Iframe.ts
```diff
@@ -0,0 +1,34 @@
...
+      ['prompt', 'none'],
+      ['scopes', this._params.scopes],
+      ['state', 'badstate']
+    ];
+
+    if (this._params.redirectURI) {
+      params.push(['redirect_uri', this._params.redirectURI]);
+    }
+
+    const url = CARTO_AUTHORIZATION_BASE + `/authorize?${buildParams(params as Array<Pair<string>>)}`;
```
VictorVelarde:
```suggestion
    const url = `${CARTO_AUTHORIZATION_BASE}/authorize?${buildParams(params as Array<Pair<string>>)}`;
```

packages/auth/src/OAuth.ts
```diff
@@ -1,12 +1,166 @@
...
+    }
+  }
+
+  public login() {
+    return new Promise((resolve, reject) => {
+      const cb = (error?: Error, token?: SalteAuth.EventWrapper) => {
+
+        if (error) {
+          reject(error);
+          this._emitter.emit('error');
```
VictorVelarde: Are we keeping this dual approach in other places, reject + event?

packages/auth/src/OAuth.ts
```diff
@@ -1,12 +1,166 @@
...
+  private _scheduleRefresh() {
+    clearTimeout(this._refreshTimeout);
+    this._refreshTimeout = window.setTimeout(() => {
+      this._refresh();
+    }, this._carto.expiresIn - Date.now() - THRESHOLD);
+  }
+
+  private _refresh() {
+    this._refresher.refresh().then((data) => {
+      // This is nasty, using salte-auth 'internals'
```
VictorVelarde: I guess there's no other way. But maybe it could be hidden after a public 'refresh' method in our CartoProvider

packages/auth/src/OAuth.ts
```diff
@@ -1,12 +1,166 @@
...
+import { Iframe } from './Iframe';
+import UserInfo from './UserInfo';
+import { unknownScopes } from './utils';
+
 class OAuth {
-  private _clientId: string;
+  private _client: SalteAuth;
+  private _carto: CartoProvider;
+  private _refresher: Iframe;
+  private _refreshTimeout: number = -1;
```
VictorVelarde: How about introducing a CONST?

packages/auth/src/OAuth.ts
rjimenezda: Well, it's complicated.

On the one hand, we are using an event system because refreshes happen completely asynchronously, and the consumer of the library cannot not have a Promise for that. Or at least, I don't see a nice way.

On the other hand, certain apps will most likely want a login.then().catch() approach. For instance, the one we're building this for.

packages/auth/tsconfig.json
```diff
@@ -12,7 +12,7 @@
     ],
     "moduleResolution": "node",
     "module": "es2015",
-    "target": "es5",
+    "target": "es6",
```
VictorVelarde: Should this change be applied to other packages?

packages/auth/src/OAuth.ts
rjimenezda: Hmm, maybe we should move all of this to CartoProvider and just call $validate from outside, or a proxy method to that.

packages/auth/tsconfig.json
rjimenezda: I don't remember exactly what the problem was, but I think a dependency was using ES6. I'd do it on a per-package basis, but I don't really mind.

When we smoke test this across several browsers I'm sure we'll find issues

packages/auth/src/OAuth.ts
VictorVelarde: Yeah, I guess we should move some responsability into it, somehow

packages/auth/src/CartoProvider.ts
rjimenezda: I'll also add an `expired` getter, I don't like checking for -1 on the main class

packages/auth/src/OAuth.ts
rjimenezda: Sure, but this is just TS bull:hankey:, the -1 has no significance. 

I have to initialize the refreshTimeout to something, and since timeoutIDs are [positive integers](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Return_value), -1 is a good fit.

VictorVelarde: ok, and how about that one below?
```
    if (this._carto.expiresIn !== -1) {
      this._scheduleRefresh();
    }
```

PR #7, opened 2019-09-25
Remove auth dependency from maps + small refactor

VictorVelarde:
After some discussion we have decided to keep 'auth' package tied to the browser, instead of a more complex isomorphic approach, so we're removing the dependency


packages/maps/src/index.ts
```diff
@@ -1,21 +1,42 @@
...
-  public instantiateMap(options: { sql?: string, dataset?: string }) {
+  public get serverURL(): string {
+    return this._serverURL;
+  }
+
+  public instantiateMap(options: any) {
+    const mapParameters = this.getMapParameters(options);
+    const requestOptions = this.getRequestOptions(mapParameters);
+
+    return this.fetchLayerGroup(requestOptions).then(this.getVectorTileURL);
```
jesusbotella: Is this working OK?

As we're using `this.getVectorTileURL` function's reference and using this within the function, I think it would fail :(

packages/maps/src/index.ts
```diff
@@ -32,18 +53,31 @@ class Maps {
...
 
-    const { username, apiKey } = this._credentials;
+  private getVectorTileURL(layerGroup: any){
+    return `${this.apiBaseUrl()}/${layerGroup.layergroupid}/{z}/{x}/{y}.mvt?api_key=${this.apiKey}`;
+  }
 
-    return fetch(`https://${username}.carto.com/api/v1/map?api_key=${apiKey}`, requestOptions)
-      .then((response) => response.json())
-      .then((data) => `https://${username}.carto.com/api/v1/map/${data.layergroupid}/{z}/{x}/{y}.mvt?api_key=${apiKey}`);
+  private apiBaseUrl() {
```
jesusbotella: What about an organization user? I think that API's work without the user's organization URL, but just in case :)

VictorVelarde:
Thx @jbotella, I'm gonna properly test it in another PR, so we can merge this without the refactor

VictorVelarde:
Ups, wrong destiny. Fixed merging into master at https://github.com/CartoDB/toolkit/pull/8

packages/maps/src/index.ts
VictorVelarde: 🤦‍♂

PR #11, opened 2019-09-30
SQL package

rjimenezda:
This PR adds a SQL package, and a top level example* that combines OAuth + SQL.

The SQL package features:

- A RequestManager, to be used per endpoint that shares rate limits. It handles limits and replays requests that fail due to them.
- A QueryManager, which is just a RequestManager that lets you do arbitrary requests.
- A CopyToManager, which is not very helpful for kepler, but might as well have it.
- A CopyFromManager, to import large files into the platform. Right now it doesn't do any file-chopping, we should try out if it works fine before complicating things. 
- All of the above, but hidden from the public interface, which has simple methods to do queries, copies, and other SQL operations.
- GRANT SELECT capabilities, we could move this to its own class in the future for finer support, but the truth is it might end up being a proper API and getting its own package so :man_shrugging: 

There's a publicly exposed static 'DDL' that helps you generate CREATE / DROP table statements, we can decide to hide it, or expose it as an own export.

Things that are missing that we might be able to afford doing on future PRs

- Document the public API
- Check body size for request and do POST instead of GET

*Keep in mind in order to test it you need to serve that index.html on https://localhost:5000 (notice the S on httpS). There's no easy solution in node that I know of, so I didn't provide any. I'm running http-server with a self signed certificate that I had to generate.


rjimenezda:
I might do the custom storage on this one instead of opening one against this branch

rjimenezda:
We have to fix the module exporting, it's kind of weird right now, and come up with a common approach for all.

packages/sql/__tests__/RequestManager.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  it('should require username, api key and a server', () => {
+    const rm = new RequestManager({ username: 'roman', apiKey: 'wadus', server: '{user}.wadus.com' });
+
+    expect(rm.username).toBe('roman');
+    expect(rm.apiKey).toBe('wadus');
+    expect(rm.server).toBe('roman.wadus.com');
+  });
+
+  it('should queue requests FIFO', () => {
+    const rm = new RequestManager({ username: 'roman', apiKey: 'wadus', server: '{user}.wadus.com' });
```
VictorVelarde: Small thing here: as this `rm` is not being used anywhere, tests are failing here:
`FAIL __tests__/RequestManager.spec.ts @carto/toolkit-sql:   ● Test suite failed to run`

packages/sql/src/CopyFromManager.ts
```diff
@@ -0,0 +1,28 @@
+import { Credentials } from './credentials';
+import RequestManager from './RequestManager';
+
+export class CopyFromManager extends RequestManager {
+
+  constructor(credentials: Credentials) {
+    super({ ...credentials, server: credentials.server + 'api/v2/sql/copyfrom' });
+  }
+
+  public query(q: string) {
```
VictorVelarde: It looks like `query` has the same implementation here and in `CopyToManager` so it could be moved up to RequestManager

packages/sql/src/RequestManager.ts
```diff
@@ -0,0 +1,167 @@
...
+        this._retryAfter = UNKNOWN;
+        // Technically? But this is risky.
+        this._callsLeft += 1;
+        this._scheduler();
+      }, this._retryAfter * 1000);
+
+      return;
+    }
+
+    const nRequests = this._callsLeft !== -1 ?
```
VictorVelarde: I would suggest a bunch of smaller methods here. It is not urgent now, but I'm gonna write them in any case, for a future refactor (after merging):

* From 60 to 66: a new small method like `private _setRetry()`
* From 71 to 84--> `private getRequestPromises()`
* From 86 to 93 --> another method to deal with previous promises and the queue (not sure about the naming)

packages/sql/src/RequestManager.ts
```diff
@@ -0,0 +1,167 @@
...
+    for (let i = 0; i < nRequests; i++) {
+      this._fetching = true;
+
+      promises.push(
+        this._fetch(this._queue[i], i)
+      );
+    }
+
+    Promise.all(promises).then((value) => {
+      this._queue = this._queue.filter((_e, i) => value.indexOf(i) === -1);
```
VictorVelarde: I find this line a bit cryptic

packages/sql/src/RequestManager.ts
```diff
@@ -0,0 +1,167 @@
...
+  private _fetch(args: FetchArgs, index: number): Promise<number | undefined> {
+    const [resolve, reject, info, init] = args;
+
+    return fetch(info, init)
+      .then((response) => {
+
+        this._retryAfter = this._getRateLimitHeader(response.headers, 'Retry-After', this._retryAfter);
+        this._callsLeft = this._getRateLimitHeader(response.headers, 'Carto-Rate-Limit-Remaining', this._callsLeft);
+
+        if (response.status === 429) {
```
VictorVelarde: Let's add a small const here about LIMITS or TIMEOUT?

packages/sql/src/credentials.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface Credentials {
```
VictorVelarde: Credentials should probably evolve to a Class in the future. The type + some methods (eg. creating a serverlURL from the original serverTemplate (with {user}) is done exactly in the same way in maps-api. 

Should not we consider a shared toolkit-core or utils?

packages/sql/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_SERVER = 'https://{user}.carto.com/';
```
VictorVelarde: Also potentially reusable from maps-api

packages/sql/src/SQLManager.ts
```diff
@@ -0,0 +1,10 @@
+import RequestManager from './RequestManager';
```
VictorVelarde: Is this class still being used? It looks like it is not anymore

index.html
```diff
@@ -0,0 +1,87 @@
+<html>
+
+<head>
+  <title>SQL tests</title>
+  <script src="packages/auth/dist/auth.umd.js"></script>
+  <script src="packages/sql/dist/sql.umd.js"></script>
```
VictorVelarde: If this example is using both packages together, should it be inside the 'toolkit' folder (like other index.html examples)?

index.html
```diff
@@ -0,0 +1,87 @@
...
+      sqlBtn.addEventListener('click', async () => {
+        const data = await copyFile(1e6);
+
+        console.log(data);
+      });
+    }
+
+    const oauth = new cartoToolkitAuth.OAuth({
+      clientID: 'fmka4wCAclcn',
+      redirectURI: 'https://localhost:5000',
```
VictorVelarde: How does anyone set this local https for a quick test? Some notes in the corresponding README?

packages/sql/src/CopyFromManager.ts
rjimenezda: Oh this is a pointless method, I'll remove it

packages/sql/src/Client.ts
```diff
@@ -0,0 +1,99 @@
...
+import { QueryManager } from './QueryManager';
+
+export class SQL {
+  private _copyToManager: CopyToManager;
+  private _queryManager: QueryManager;
+  private _copyFromManager: CopyFromManager;
+  private _publicQueryManager: QueryManager;
+  private _publicRole?: string;
+
+  constructor(username: string, apiKey: string, server: string = DEFAULT_SERVER) {
```
VictorVelarde: should we use Credentials here too? Having an object could be useful to add more (optional) properties in the future.

I'm currently using (at maps-api) `constructor(username: string, apiKey: string, serverURL?: string, client?: string)` 

A client set to  'toolkit-maps' / 'toolkit-sql' could be useful to get some metrics

packages/sql/src/RequestManager.ts
rjimenezda: Every promise returns its index on the queue when it finishes correctly (not if it needs a retry because of 429), when they're all done, we filter the queue to remove those that properly resolve.

I think it's extra cryptic because `value` should be `finishedPromises` or something

packages/sql/src/RequestManager.ts
rjimenezda: TOO_MANY_REQUESTS would be the most appropriate

packages/sql/src/SQLManager.ts
rjimenezda: Nope, it's an early test

packages/sql/src/credentials.d.ts
rjimenezda: Yeah, I thought the same, but I don't how to solve this.

Keep in mind making it a separate package means we have to publish it on npm, and it would be the most pointless package ever (and there's lots of competition).

We could have a "private" package (so, a folder, not really a dependency) that we include, but I don't know if that is possible with TS.

However, we have to solve the credentials issue, because I think the natural way is to pass an object instead of a string, and just read the apiKey from there every time, so if it updates externally (i.e: OAuth), it will just work.

index.html
rjimenezda: IMHO, not really. The `toolkit` folder should use the `toolkit` package which will hide all of these packages behind a common interface.

This is just a smoke test to be honest.

index.html
rjimenezda: There's a few examples around. Really, what you have to do, is figure out how to generate a cert/pem/key on you OS, then you can use [`http-server`](https://www.npmjs.com/package/http-server) or any other.

It's kind of annoything, react-cli and vue-cli have http servers that generate a self signed certificate for you, but it's not on a ready to use package.

Zeit's serve used to do this, but it no longer has that feature

packages/sql/src/Client.ts
rjimenezda: I think we should use something else, probably something that the auth package can provide, but not necessarily as a real dependency.

An object with the shape of those arguments would work.

I think we should tackle this after we're done, and solve the issue of updating the API key by adding a setter _for now_

VictorVelarde: yeah, for sure, let's move on. I just wanted to point out the need of thinking more about this common pieces / repetitions

Issue #15, opened 2019-09-30
Toolkit UMD naming


rjimenezda:
The UMD module names are a bit of a mess right now. They are long, verbose, and not namespaced.

We should try to come up with a relatively short way of naming packages, maybe even namespacing?

IDEA: We could have different entry points for the UMDs, so they build a namespace. We shouldn't use `carto.` because I think VL and carto.js will not play nice.

VictorVelarde:
At this point in time, I think we should go for a new namespace, starting from 'carto' root.

Something in the line of:
- carto.vis: new viz classes to come
- carto.data (<-- sql): SQL API and more to come
- carto.auth: Auth / OAuth stuff
- carto.maps: Maps API
- carto.customStorrage (maybe just rename it to carto.kepler package?)

Opening the discussion with you @jesusbotella, @neokore and @elenatorro

rjimenezda:
As long as it doesn't clash with VL and carto.js, which also live under 'carto'

VictorVelarde:
or even if it does... 😄 (more to come soon...)

neokore:
Taking `carto` as root is great, it makes easy to remember and sticks the library with the brand, but depending on how it will be distributed, we may need to add some "legacy users" checks.
E.g. in case of reusing a CARTO package (let's say carto.js) and any user just upgrade it, it will be useful to detect legacy library calls (maybe the initialize methods) and throw an exception to alert that this library is not backwards compatible.

jesusbotella:
I think we should use `carto` namespace no matter what.
The library/module is not intended to work alongside CARTO-VL or CARTO.js, they have their own modules starting from `carto` namespace. So no need to be backwards compatible with our own libraries.

VictorVelarde:
We'll make the move just after merging metrics-event PR #56 , starting a new base 'viz', isolated from the current -rc.013 line

PR #16, opened 2019-10-01
Feature/custom storage

rjimenezda:
First version of a "Custom Storage Engine", basically tailored for what we need on Kepler.

The idea here is to use the SQL API extensively to:

- Have two tables for custom visualizations, one for private, one for public ones. The public one gets `grant select on <name>`.
- Have a table related to those to store which datasets are on which visualizations.
- The public one does some things with the master key, some with `default_public`, so it has two SQL clients under the hood.

The Custom Storage Engine takes care of all the heavy lifting: creating, uploading using COPY, inserting into the appropriate tables. Same on the other way around, when deleting stuff.

Missing from this PR:

- Cartodbfiying tables, although the procedure is a simple transaction, documented on the code. The trick will be calling the `/me` API so we can call cdb_cartodbify properly. 
- Updating a visualization: this is not required for the public map sharing, so we can do it later. Neither is deleting, but I did that one.
- Figure out a better way to create the UUID function. Because it's part of the DDL of the tables, you can't DROP it each time, and we're creating it twice.

Refactors required:
- Improve SQL API return types, I've had to type too many `:any`, we should be able to say it might have an error, and has a `rows` field.


rjimenezda:
Also, I've switched to ES6 because otherwise async / await doesn't work as you'd expect apparently.

For kepler that's fine, it would only exclude IE11 which is not very well supported I think. We _could_ add a babel transpiler if we think it's important.

index.html
```diff
@@ -26,35 +27,42 @@
...
+        contents.push(`${i},A`);
       }
 
-      async function copyFile(rows = 7) {
-        await sql.query('truncate copy_test');
+      return contents.join('\n');
+    }
+
+    async function copyFile(rows = 7) {
+      await sql.create('copy_test', ['id numeric', 'name text']);
```
VictorVelarde: should this have a 'ifNotExists'?

index.html
```diff
@@ -26,35 +27,42 @@
...
 
-        return data;
-      }
+      const data = await sql.query('SELECT * FROM copy_test LIMIT 100');
+
+      return data;
+    }
+
+    async function init(token) {
+      sql = new cartoToolkitSql.SQL(
```
VictorVelarde: We should probably agree on all these 'rollup' names for umd, as they are going to be the 'public face' of the library outside, and it would be hard to change them in the future

packages/custom-storage/src/StorageRepository.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface CompleteVisualization {
+  vis: StoredVisualization;
+  datasets: Dataset[];
+}
+
+export interface Visualization {
+  name: string;
+  desc: string;
+  img: string;
```
VictorVelarde: why not a full 'description' and 'image' here?

PD. I wanted to use the new 'multi-line comments' for sure...  😛

packages/custom-storage/src/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEFAULT_SERVER = 'https://{user}.carto.com/';
+export const DEFAULT_API_KEY = 'default_public';
```
VictorVelarde: Yeah, that 'common' again

packages/custom-storage/src/StorageRepository.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface StorageRepository {
+  getVisualizations(): Promise<StoredVisualization[]>;
+  getPublicVisualizations(): Promise<StoredVisualization[]>;
+  getPrivateVisualizations(): Promise<StoredVisualization[]>;
+  getVisualization(id: string): Promise<CompleteVisualization | null>;
+  getPublicVisualization(id: string): Promise<CompleteVisualization | null>;
+  deleteVisualization(id: string): Promise<boolean>;
+  createVisualization(visualization: Visualization, datasets: Dataset[]): Promise<any>;
+  updateVisualization(visualization: StoredVisualization, datasets: Dataset[]): Promise<any>;
```
VictorVelarde: A general question about 'StoredVisualization' and other interfaces referring to different states. 

What's the main advantage you see over using just a plain class `Visualization` with all possible fields, some of them null or empty and some check XXX methods?

index.html
rjimenezda: https://github.com/CartoDB/toolkit/issues/15

packages/custom-storage/src/index.ts
```diff
@@ -0,0 +1,103 @@
...
+    return this._privateSQLStorage.getVisualizations();
+  }
+
+  public getVisualization(id: string): Promise<CompleteVisualization | null> {
+    // Alternatively: SELECT * from (SELECT * FROM <public_table> UNION SELECT * FROM <private_table>) WHERE id = ${id};
+    return Promise.all([
+      this._publicSQLStorage.getVisualization(id),
+      this._privateSQLStorage.getVisualization(id)
+    ]).then((d) => {
+      return d[0] || d[1];
```
VictorVelarde: Nice hacky detail!

packages/custom-storage/src/StorageRepository.ts
rjimenezda: Less code and less state management, mainly.

Also I think that, at a glance, you can see that they're not exactly the same. If both signatures were visualization: Visualization, you might think that you can specify the id on `create`, when you really shouldn't. However, you absolutely need to have it on `update`

packages/custom-storage/src/StorageRepository.ts
rjimenezda: Good question, I don't know

Actually, `desc` is an issue with SQL, so initially it was also `desc` for the DDL, but I had to change it. Also, `img` is `thumbnail` on the DB, which is a big inconsistency.

packages/sql/src/QueryManager.ts
```diff
@@ -1,17 +1,28 @@
...
 
-  public query(q: string) {
+  public query(q: string, extraParams: Array<Pair<string>> = []) {
+    const urlParams = [
+      ['api_key', this.apiKey],
+      ['q', q],
+      ...extraParams
+    ];
+
+    const stringParams = encodeURI(urlParams.map(
```
VictorVelarde: again, very similar set of encoding ops to the ones in maps-api

packages/custom-storage/src/constants.ts
rjimenezda: Just tested this, looks like you can have a common folder you can import, so we could put some constants, utils and classes in there.

rjimenezda: Actually, no, all source code must have the same common top folder.

VictorVelarde: That's cool! We can talk about it tomorrow.

rjimenezda:
Sorry, I had to force push

packages/custom-storage/src/StorageRepository.ts
VictorVelarde: Let's try to keep them homogeneous then

rjimenezda: Already fixed

packages/custom-storage/src/constants.ts
VictorVelarde: Let's keep the conversation outside of this PR --> https://github.com/CartoDB/toolkit/issues/17

packages/custom-storage/src/StorageRepository.ts
VictorVelarde: Ok, let's see how that works 👌

PR #19, opened 2019-10-02
Feature/metapackage

rjimenezda:
This PR adds the code to the 'toolkit' package, which exposes all other packages and creates the concept of an 'App'. This package serves a twofold purpose:

- It lets you install it and use each package individually
- Use the App / OAuthApp that will do all the internal bindings for you.

Besides the code itself, I'd appreciate some input on the how the export looks right now. I've standardized it (check out how @carto/toolkit imports them, and the index.html).

Also, App / OAuthApp are on the same level as SQL, OAuth, etc...should we create a level for them? CartoApp? I dunno


rjimenezda:
Note to self: after the amazing async refactor, the first time you log into an OAuth app, the getters won't work, but they do if you refresh. Must fix this, it's a bug present on rc.4

Issue #5, opened 2019-09-23
Add CI configuration


VictorVelarde:
Prepare a simple CI integration. 

CircleCI worked pretty well in other projects for us

rjimenezda:
Fixed via #20

PR #20, opened 2019-10-04
Add CI with CircleCI

VictorVelarde:
To deal with https://github.com/CartoDB/toolkit/issues/5


.circleci/config.yml
```diff
@@ -0,0 +1,45 @@
...
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+          - cache-01-dependencies-{{ checksum "package-lock.json" }}
+          - cache-01-dependencies-
+      - run: npm install
+      - run: npm run bootstrap
+      - run: npm run lint
```
rjimenezda: I just remembered, we don't really have this on the subpackages I think

PR #22, opened 2019-10-08
Feature/read credentials from Apps

### LINKED ISSUE #21: [toolkit] Add getter to read back credentials
For kepler, we need to know what user is logged in via OAuth.

We should do this on the base class, which, by the way, should not allow setting credentials more than once.

rjimenezda:
Fix #21

### DIFF
--- a/packages/toolkit/index.html
+++ b/packages/toolkit/index.html
@@ -8,7 +8,6 @@
 <body>
   <div id="online">
     <button id="logoutBtn">Logout</button>
-    <button id="sqlBtn">Truncate, copy 1e6 rows and select 1000</button>
   </div>
   <div id="offline">
     <button id="loginBtn">Login</button>
@@ -57,7 +56,9 @@
         USERNAME
       );
 
-      (await masterApp.CustomStorage).destroy();
+      const cs = await masterApp.getCustomStorage();
+      
+      cs.destroy();
     }
 
   </script>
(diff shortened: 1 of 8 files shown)


packages/toolkit/src/App.ts
```diff
@@ -69,6 +76,16 @@ class App {
     return this._initPromise;
   }
 
+  public setApiKey(apiKey: string) {
```
jesusbotella: Is this meant for public use?
This method does not follow return convention on `setCredentials` method of returning a promise, and it doesn't return an error when SQL or CustomStorage is not available.

packages/toolkit/index.html
```diff
@@ -57,7 +57,7 @@
         USERNAME
       );
 
-      masterApp.CustomStorage.destroy();
+      (await masterApp.CustomStorage).destroy();
```
jesusbotella: Awaiting a variable seems strange to me, but I understand the why though. Shouldn't App class act as a façade to all those actions?

packages/toolkit/src/App.ts
rjimenezda: Eh, we could make it protected. It's a method to update the token when it's updated on OAuth apps, there's no point in allowing changing the token from outside.

I didn't return the promise because this has no async consequences, but you're probably right that we should be consistent with the asynchronicity.

packages/toolkit/index.html
rjimenezda: I also find async getters weird, when it started as a getter it was not async, but it evolved.

Maybe we should make it a `getCustomStorage` instead?

PR #34, opened 2019-10-23
Minify packages

neokore:


lerna.json
```diff
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.1-rc.8"
+  "version": "0.0.1-rc.9"
```
jesusbotella: Not a major issue now, but we tend to bump versions whenever they're released so that we can group several changes into a single version if we need it.

rjimenezda: I actually have been doing RC's on branches this whole time. But yeah, proper versions we have a different strategy.

This looks like he tried to prerelease and it failed (he probably cannot release on npm) but the commit and tag remain there.

neokore:
Created `build:debug` generation adding minify/uglify plugins (`terser` so far) to an array in order to apply it only when needed without creating different configs.
Also enabled sourcemaps to TypeScript to fix empty sourcemaps issue.

Issue #30, opened 2019-10-17
[SQL] Improve error handling


rjimenezda:
Right now, rate limits are handled gracefully. However, we should also be wary of:

- 503 errors, that might happen if the SQL API is not ready for whatever reason and nginx can't redirect the request to it. We should just retry after some time (maybe immediately)
- Timeout errors, these might come back as a 429, but probably don't have the retry headers. To try this, PG_Sleep is probably the way.

jesusbotella:
Do we have any retry limit? I think that if we retry 503 errors right away we're gonna end up in a loop.

jesusbotella:
Merged!

PR #35, opened 2019-10-24
Retry requests when backend service is unavailable or times out

jesusbotella:
This PR implements phased request replay when backend service is unavailable or there's a timeout error not related to rate limits (query takes too long).

I think we can hit some race conditions that would lead to retrying beyond attempts limit, but let's leave it like it is.

Related to #30

### DIFF
--- a/packages/sql/src/RequestManager.ts
+++ b/packages/sql/src/RequestManager.ts
@@ -71,6 +71,11 @@ export class RequestManager {
     }
 
     if (this._retryAfter !== UNKNOWN) {
+      // This timeout waits for the minimum time to
+      // call scheduler and reset previous retry values
+      // It adds 1 to calls left because as we've waited for the
+      // limit to expire there should be one call available at least
+      // without taking into account other possible sources of requests
       this._retryTimeoutId = window.setTimeout(() => {
         this._retryTimeoutId = UNKNOWN;
         this._retryAfter = UNKNOWN;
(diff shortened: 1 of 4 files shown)


packages/sql/src/RequestManager.ts
```diff
@@ -99,28 +108,41 @@ export class RequestManager {
...
-          // Reschedule
-          this._scheduler();
-          return null;
+        const responseBody = await getResponseBody(response);
+
+        const isTimeoutError = response.status === HTTP_ERRORS.TOO_MANY_REQUESTS &&
+          responseBody.detail === 'datasource';
+
+        if (response.status === HTTP_ERRORS.SERVICE_UNAVAILABLE || isTimeoutError) {
+          requestDefinition.retries_no = retries_no !== NO_RETRY ? retries_no - 1 : MAX_RETRIES;
```
neokore: In case of `retries_no` is equal to `NO_RETRY`, we are setting `retries_no` to `MAX_RETRIES`. Shouln't it be set to `NO_RETRY` in this case?

packages/sql/src/RequestManager.ts
```diff
@@ -99,28 +108,41 @@ export class RequestManager {
...
+
+          const timeToWait = (MAX_RETRIES - requestDefinition.retries_no) * 0.5 + 0.5;
```
rjimenezda: I'd move these `0.5` to some constants like RETRY_FACTOR and RETRY_MIN_WAIT or something

packages/sql/src/RequestManager.ts
```diff
@@ -99,28 +108,41 @@ export class RequestManager {
     });
   }
 
-  private _fetch(args: FetchArgs, index: number): Promise<number | undefined> {
-    const [resolve, reject, info, init] = args;
+  private _fetch(requestDefinition: FetchArgs, index: number): Promise<number | undefined> {
+    const {resolve, reject, requestInfo, requestInit, retries_no} = requestDefinition;
```
rjimenezda: Much better

packages/sql/src/RequestManager.ts
rjimenezda: This is pedantic, but calling it `retries_no` and then having `NO_RETRY` makes this weird to read. We could rename it to `retries_count`or something.

jesusbotella:
I've changed review changes proposals and added two comments 😁 

Let me know if you want me to add some other comments :)

Issue #36, opened 2019-10-25
[toolkit] Add Last modified field to kepler viz


VictorVelarde:
So it allows a better integration in the Dashboard with Builder maps

VictorVelarde:
Included in PR for 0.1

PR #37, opened 2019-10-28
Add last_modified field to visualization model

jesusbotella:
Related to https://github.com/CartoDB/toolkit/issues/36


packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+describe('SQLStorage', () => {
+  let sqlStorage: SQLStorage;
+
+  beforeAll(() => {
+    const sqlClient = new SQL('user', 'm14p1k3y', 'https://{user}.carto.com/');
+    sqlStorage = new SQLStorage('mynamespace', sqlClient, 1, true);
+  });
+
+  it('should check or create tables for storage', async () => {
```
VictorVelarde: I would suggest adding "on init"

packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -0,0 +1,179 @@
+import { SQL } from '@carto/toolkit-sql';
+import { SQLStorage } from '../src/sql/SQLStorage';
+import {
+  StoredVisualization,
+  // Visualization,
+  // Dataset
+} from '../src/StorageRepository';
+import {
+  CHECK_CREATE_QUERIES,
```
VictorVelarde: Nice!
I sometimes do that kind of extraction, specially if the queries are going to be reused. If they are not, they could also be inside the test itself (they are not so long after all...), to avoid an indirection, but that's just my preference

packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    // Check the creation and access grant queries of the 3 tables
+    (global as any).fetch.mock.calls.forEach((call: any, idx: number) => {
+      const params = (global as any).getURLParams(call[0]);
+      expect(decodeURI(params.q)).toBe(CHECK_CREATE_QUERIES[idx]);
+    });
+  });
+
+  it('should list visualizations', async () => {
+    const visObject: StoredVisualization = {
```
VictorVelarde: Maybe we can extract the content into a DATA const, to reuse both here and in the mock

neokore: :+1:

packages/custom-storage/__tests__/SQLStorage.spec.ts
neokore: Totally right, I'll change it!

packages/custom-storage/__tests__/SQLStorage.spec.ts
neokore: Sure!

packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  beforeAll(() => {
+    const sqlClient = new SQL('user', 'm14p1k3y', 'https://{user}.carto.com/');
+    sqlStorage = new SQLStorage('mynamespace', sqlClient, 1, true);
+  });
+
+  it('should check or create tables for storage', async () => {
+    // We use only one mock because is not important for the rest of the queries
+    (global as any).fetch.mockResponse(
+      JSON.stringify({
```
VictorVelarde: What do we use this response for in the test?

Aren't we more interested in the queries sent themselves than in the response?
I was wondering if spying the SQL calls would have been enough, instead of replacing fetch

VictorVelarde: What do you think @jesusbotella ?

packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const storedVis = await sqlStorage.getVisualizations();
+
+    // Check list visualizations query
+    const lastCall = (global as any).fetch.mock.calls[(global as any).fetch.mock.calls.length - 1];
+    const params = (global as any).getURLParams(lastCall[0]);
+    expect(decodeURI(params.q)).toBe(LIST_VISUALIZATION_QUERY);
+
+    // Check response
+    expect(storedVis).toStrictEqual([visObject]);
```
VictorVelarde: Here, in the second expect we are probably testing 'rowToVisualization' from utils, more than the SQLStorage. getVisualizations(). I think that the first check could probably be enough

packages/custom-storage/__tests__/SQLStorage.spec.ts
neokore: This response is used for granting access to the tables (the last 3 queries). If this data is not added on response, the method will fail with a `Cannot grant: got no current_user data` exception.

packages/custom-storage/__tests__/SQLStorage.spec.ts
neokore: Yes, this tests could fit better for `rowToVisualization`; but I thought that it would be important to always check the returned value of a function to avoid issues related to changing its format. Anyway, I can move this test to an utils file if you prefer.

packages/custom-storage/__tests__/SQLStorage.spec.ts
VictorVelarde: Thx, now I see it more clearly!; so the rolename determines the SQL sentences being generated. Perhaps having them inline, using that same rolename (as a shared variable?) could be more clear.

PR #42, opened 2019-11-06
Temporary remove foreign keys

neokore:
Related to #43

### DIFF
--- a/packages/custom-storage/src/sql/SQLStorage.ts
+++ b/packages/custom-storage/src/sql/SQLStorage.ts
@@ -56,8 +56,9 @@ export class SQLStorage {
     ];
 
     this.DATASET_VIS_COLUMNS = [
-      `vis uuid references ${this._tableName}(id) ON DELETE CASCADE`,
-      `dataset uuid references ${this._datasetsTableName}(id) ON DELETE CASCADE`
+      // TODO: Reenable foreign keys when https://github.com/CartoDB/cartodb/issues/15161 is solved
+      `vis uuid NOT NULL`, // `vis uuid references ${this._tableName}(id) ON DELETE CASCADE`,
+      `dataset uuid NOT NULL` // `dataset uuid references ${this._datasetsTableName}(id) ON DELETE CASCADE`
     ];
 
     this.FIELD_NAMES = (Object.values(this.VIS_FIELDS) as ColumConfig[])
--- a/packages/custom-storage/src/sql/SQLStorage.ts
+++ b/packages/custom-storage/src/sql/SQLStorage.ts
@@ -100,8 +100,8 @@ export class SQLStorage {
     return getVisualization(this._tableName, this._datasetsTableName, this._datasetsVisTableName, id, this._sql);
   }
 
-  public async getDatasetData(name: string): Promise<Dataset> {
-    return getDatasetData(name, this._sql);
+  public async getDatasetData(name: string, tablename: string): Promise<Dataset> {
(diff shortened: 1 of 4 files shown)


neokore:
Synced with master, not it should pass tests ;)

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -143,6 +144,10 @@ export class SQLStorage {
   }
 
   public async deleteVisualization(id: string): Promise<void> {
+    // TODO: Delete dataset if is not used by any other visualization
```
VictorVelarde: I think we can try to add it now, to avoid orphans.

Maybe we can go on something like:
```sql
DELETE 
FROM keplergl_public_v0_datasets 
WHERE id NOT IN 
	(SELECT distinct(dataset) 
     FROM keplergl_public_v0_datasets_vis
    );
```

Adding that as the last query, once the record in the intermediate table has been deleted.

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -318,6 +323,7 @@ export class SQLStorage {
 
     const datasets: string[] = rawDatasets.map((row: { name: string }) => row.name);
 
+    // NOTE: DROP TABLE CASCADE removes dependant views or functions and foreign keys constraints (neither tables nor data)
     return this._sql.query(`
       BEGIN;
         ${datasets.map((datasetName) => `DROP TABLE IF EXISTS ${datasetName};`).join('\n')}
```
VictorVelarde: I didn't know we were removing the datasets themselves! It's ok by now if this is used just for us know, but it looks a bit dangerous to me. I guess we should start to document at least our methods, describing this set of deletions

packages/toolkit/examples/oauthApp.html
```diff
@@ -218,6 +222,11 @@ <h4>Debug Output</h4>
 
       output(result);
     });
+
+    deleteVisBtn.addEventListener('click', async () => {
```
VictorVelarde: It's nice to have a small example here

VictorVelarde:
It looks nice to me. I just added the delete orphan datasets suggestion. Once you consider that you can go ahead and merge

VictorVelarde:
If we're going to bump a new toolkit version, we can take the change to fix first the 
```
// TODO: customStorage should return dataset name without prefix
const datasetId = dataset.name.split('keplergl_public_v0_')[1];
```

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -100,8 +100,8 @@ export class SQLStorage {
     return getVisualization(this._tableName, this._datasetsTableName, this._datasetsVisTableName, id, this._sql);
   }
 
-  public async getDatasetData(name: string): Promise<Dataset> {
-    return getDatasetData(name, this._sql);
+  public async getDatasetData(name: string, tablename: string): Promise<Dataset> {
```
jesusbotella: Given that name was used as the table name previously, why don't we change the parameter name?

I don't get how we are stripping the prefix from table's name 🤔

neokore: It's not really needed to split table's name, we are getting both dataset name and table name when getting dataset info; so we just pass it as a parameter to keep it simple, as it is done in `custom-storage/src/sql/utils.ts@78`.

Issue #52, opened 2020-01-14
[customStorage] Issues on updating and deleting visualizations


neokore:
- [x] Fix repeated `datasets_vis` entries when saving an existing visualization
- [x] Review deleting old datasets when saving an existing visualizations

~- [ ] Allow configure retry management in requests queue (set max retries value)~

VictorVelarde:
Let's split this up, keeping bux fixing on tables separated from improvements.

VictorVelarde:
I guess this should be closed now, @neokore. Feel free to reopen if needed

PR #53, opened 2020-01-16
Update custom storage visualizations feature

neokore:
- `updateVisualization` method implemented.
- ~⚠️ Note that although a visualization is unlinked to a dataset, the related dataset is not deleted.~
- Some column names were changed due to conflicts with reserved names.

Related to https://github.com/CartoDB/toolkit/issues/52


packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('SQLStorage', () => {
 
   it('should check or create tables for storage on init', async () => {
     const expectedQueries = [
-      'CREATE TABLE IF NOT EXISTS mynamespace_public_v1 (id uuid PRIMARY KEY DEFAULT mynamespace_create_uuid(), name text NOT NULL, description text, thumbnail text, private boolean, config json, last_modified timestamp NOT NULL DEFAULT now());',
+      'CREATE TABLE IF NOT EXISTS mynamespace_public_v1 (id uuid PRIMARY KEY DEFAULT mynamespace_create_uuid(), name text NOT NULL, description text, thumbnail text, isprivate boolean, config json, lastmodified timestamp NOT NULL DEFAULT now());',
```
VictorVelarde: I now about the private 'reserved' word issue, but out of curiosity, why did you change 'last_modified'?

VictorVelarde:
I have not tested it but I think we need a bit of a refactor, cause insert & update share a lot of code. Let me give it a try

VictorVelarde:
I made a small refactor. I hope everything's fine regarding the new names. 

We have to discuss a bit about the orphan datasets. I would say we should probably delete them too.

packages/custom-storage/__tests__/SQLStorage.spec.ts
neokore: This was a quick fix for a mismatch between `lastModified` in object and `last_modified` in the DB; to avoid a deeper change when we wanted to finish v0.2 soon.
But thinking on it again, it could be better to create another field in columns for `nameDB` or just create an object adapter to transform between JS and DB.

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -161,85 +164,19 @@ export class SQLStorage {
...
-        await this.shareDataset(tableName);
-      }
-    }
-
+    await this.uploadAndLinkDatasetsTo(insertedVis.id, datasets, overwriteDatasets, vis.isPrivate);
 
     return {
-      id,
-      lastModified,
+      ...insertedVis,
```
neokore: I think we need to change this order to avoid `...vis` to overwrite values from `insertedVis`

VictorVelarde: Do we? This is on creation, so 'id' and 'lastmodified' shouldn't be set on the 'vis' yet...

neokore: Sorry, I mean on update

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -297,72 +234,18 @@ export class SQLStorage {
...
-      // We'll need some extra user info for this step, fetch this early on.
-      // CARTODBFY(...);
-      // END;
-    }
+    await this.uploadAndLinkDatasetsTo(updatedVis.id, datasets, true, vis.isPrivate);
 
     return {
-      id,
-      lastModified,
+      ...updatedVis,
```
neokore: Here is where we should swap variables

packages/custom-storage/__tests__/SQLStorage.spec.ts
VictorVelarde: Feel free to improve the 'FIELDS' with any mechanism you think is better to avoid mistakes

packages/custom-storage/src/sql/SQLStorage.ts
VictorVelarde: Yes, it looks so. 👌

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -350,6 +290,120 @@ export class SQLStorage {
...
+    RETURNING ${this.VIS_FIELDS.id.name}, ${this.VIS_FIELDS.lastModified.name}
+  `);
+
+    if (updatedResult.error) {
+     return null;
+   }
+
+    return {
+     id: vis.id,
+     lastModified: updatedResult.rows.length ? updatedResult.rows[0].lastmodified : vis.lastModified
```
VictorVelarde: I modified your `rows[0].lastModified` with `.lastmodified`. Double check it please

packages/custom-storage/src/sql/SQLStorage.ts
neokore: After testing it, it appears that it keeps last modification field, so no changes needed here.

packages/custom-storage/src/sql/SQLStorage.ts
neokore: It is right

VictorVelarde:
Ready to merge after you can have a look at it

VictorVelarde:
I have fixed the linter, merged master into and added the orphan datasets truncation (not tested yet)

packages/sql/src/RequestManager.ts
```diff
@@ -127,7 +127,7 @@ export class RequestManager {
         const responseBody = await getResponseBody(response);
 
         const isTimeoutError = response.status === HTTP_ERRORS.TOO_MANY_REQUESTS &&
-          responseBody.detail === 'datasource';
+          (responseBody.detail === 'datasource' || responseBody.detail === 'rate-limit');
```
VictorVelarde: ✨ Good catch!

PR #54, opened 2020-01-24
Drop not required package-lock files in packages + hoisting

VictorVelarde:
It looks like this is the common practice when in lerna projs, to avoid unneeded locks and optimize performance.

See for example https://github.com/CondeNast/atjson/pull/40/commits/984ea21528ceffcb8bdb223794b477cf0a40cc04


jesusbotella:
👋!

I've just read the issue, and I do agree that we don't need all those `package-lock.json` files, but we are not hoisting those `node_modules` dependencies.

I checked our dependencies, and for example, `@salte-auth` dependencies in are not logged in the main `package-lock.json`, while they are within `auth/package-lock.json` which is the module where they are listed.

So I am concerned about dependencies not being version locked :(

VictorVelarde:
It looks like running `lerna bootstrap --hoist` is enough to push them up and get hoisting effectively. eg. @salte-auth is now declared as dependency in the root level.

jesusbotella:
So, we're now declaring dependencies in our top-level package.json, right?

https://github.com/CartoDB/toolkit/blob/214d961122a9046fa24abfa3bb561d2b96dfbfcc/package.json#L9-L26

VictorVelarde:
Yes, from my understanding it allows managing the "common parts" in that level with more efficiency. A

Some notes about hoisting: https://github.com/lerna/lerna/blob/master/doc/hoist.md.

To avoid being able to import not explicit dependencies, eslint-plugin-import --> `no-extraneous-dependencies`rule is recommended. I have applied what looks like the equivalent for tslint: https://palantir.github.io/tslint/rules/no-implicit-dependencies/

packages/auth/src/OAuth.ts
```diff
@@ -24,7 +24,7 @@ class OAuth {
     args.authorization = args.authorization || CARTO_AUTHORIZATION_BASE;
     args.scopes = args.scopes || '';
 
-    this._emitter = new mitt();
+    this._emitter = new (mitt as any)();
```
jesusbotella: Doesn't it work with mitt()? What error does it throw?
Doing this is far from good 😁

package.json
```diff
@@ -1,21 +1,39 @@
...
-    "test": "lerna run test --stream",
-    "lint": "lerna run lint --stream",
-    "lint:fix": "lerna run lint:fix --stream",
-    "release": "lerna bootstrap && npm run test && npm run build && lerna publish --force-publish=* --conventional-commits",
-    "prerelease": "lerna bootstrap && npm run test && npm run build && lerna publish --dist-tag prerelease --force-publish=* --preid rc"
-  }
+	"name": "root",
+	"private": true,
+	"devDependencies": {
+		"@types/jest": "^24.0.18",
```
jesusbotella: I don't know why this changed indentation to 4 spaces 🤔

package.json
```diff
@@ -1,21 +1,39 @@
...
+		"rollup-plugin-typescript2": "^0.24.2",
+		"ts-jest": "^24.1.0",
+		"tslint": "^5.20.0",
+		"typescript": "^3.6.3"
+	},
+	"scripts": {
+		"watch": "lerna run watch --parallel --stream",
+		"build": "lerna run build --stream",
+		"build:debug": "lerna run build:debug --stream",
+		"bootstrap": "lerna bootstrap --hoist",
```
jesusbotella: This command is good, but I think we should add `postinstall` (like we have in Airship) so that we don't forget to lerna bootstrap.

packages/auth/src/OAuth.ts
VictorVelarde: After applying the hoisting, the build was broken, throwing this:

```
@carto/toolkit-auth: Error: /Users/victor/CARTO/toolkit/packages/auth/src/OAuth.ts(27,21): semantic error TS7009: 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.
```

(using tsc 3.6.3)

package.json
VictorVelarde: Yeah, that's better. Done on the next commit.

package.json
VictorVelarde: The first bootstrap hoist execution removed the 'scripts' section from the file, and I had to manually recover that, so that's probably my fault.

packages/auth/src/OAuth.ts
jesusbotella: Oh, I remember the same thing using Airship. I don't know why we use here `new` keyword, it is not needed as [documentation](https://github.com/developit/mitt#typescript) shows:

```js
import mitt from 'mitt';
const emitter: mitt.Emitter = mitt();
```

VictorVelarde: Done!

Issue #40, opened 2019-11-05
[toolkit / kepler.gl] Identify API calls from app client (Kepler.gl)


VictorVelarde:
See: https://github.com/CartoDB/product/issues/467#issuecomment-539905698

After some research and conversations, a plan has been set on **https://app.clubhouse.io/cartoteam/epic/57961/api-and-client-metrics** with different steps:

- [x] Adding headers to SQL API https://github.com/CartoDB/CartoDB-SQL-API/pull/640
- [x] Implementing events on custom-storage #56 
- [x] Check with Back 
- [x] Allow to distinguish between kepler.gl web app & dashboard

neokore:
Applied in Dashboard. Kepler.gl will follow.

VictorVelarde:
Great, let's close this when rc-17 is on prod on kepler.gl side

neokore:
Uber merged my PR and this is now live on prod on Kepler.gl!!

PR #56, opened 2020-02-03
Add MetricsEvent to core & custom-storage main events

VictorVelarde:
Deals with https://github.com/CartoDB/toolkit/issues/40 and corresponds to https://app.clubhouse.io/cartoteam/story/58905/implement-client-metrics-in-kepler-sql

These are the events to be sent to APIs (*Carto-Source-Context* values):

<details>
<summary>~OLD Naming proposal~</summary>

- **CustomStorage**:
 const CONTEXT_INIT = 'custom-storage-init';
 const CONTEXT_CREATE_VIS = 'custom-storage-create-visualization';
 const CONTEXT_UPDATE_VIS = 'custom-storage-update-visualization';
 const CONTEXT_GET_ALL_VIS = 'custom-storage-get-visualizations';
 const CONTEXT_GET_PUBLIC_VIS = 'custom-storage-get-public-visualizations';
 const CONTEXT_GET_PRIVATE_VIS = 'custom-storage-get-private-visualizations';
 const CONTEXT_GET_VIS = 'custom-storage-get-visualization';

- **PublicSQLReader**:
const CONTEXT_GET_PUBLIC_VIS = 'public-sql-reader-get-visualization';

So for example, creating a visualization from 'keplergl' will issue several requests to our APIs, with headers like:

'Carto-Source-Lib': 'kepler'
'Carto-Source-Context': 'custom-storage-create-visualization'
'Carto-Source-Context-Id': 'some_random_uuid'

</details>

AFTER some discussion with @esloho and @ilbambino, we will finally use:

- **CustomStorage**:
 const CONTEXT_INIT = 'custom_storage_init';
 const CONTEXT_CREATE_VIS = 'custom_storage_visualization_create';
 const CONTEXT_UPDATE_VIS = 'custom_storage_visualization_update';
 const CONTEXT_GET_ALL_VIS = 'custom_storage_visualization_list_load';
 const CONTEXT_GET_PUBLIC_VIS = 'custom_storage_public_visualizations_load';
 const CONTEXT_GET_PRIVATE_VIS = 'custom_storage_private_visualizations_load';
 const CONTEXT_GET_VIS = 'custom_storage_visualization_load';

- **PublicSQLReader**:
const CONTEXT_GET_PUBLIC_VIS = 'public_sql_reader_visualization_load';

So for example, creating a visualization from 'keplergl' will issue several requests to our APIs, with headers like:

Carto-Event-Source: kepler
Carto-Event: custom_storage_visualization_create
Carto-Event-Group-Id: some_random_uuidv4

### DIFF
--- a/packages/custom-storage/src/CustomStorage.ts
+++ b/packages/custom-storage/src/CustomStorage.ts
@@ -13,7 +13,7 @@ import {
 const CONTEXT_INIT = 'custom_storage_init';
 const CONTEXT_CREATE_VIS = 'custom_storage_visualization_create';
 const CONTEXT_UPDATE_VIS = 'custom_storage_visualization_update';
-const CONTEXT_GET_ALL_VIS = 'custom_storage_visualizations_load';
+const CONTEXT_GET_ALL_VIS = 'custom_storage_visualization_list_load';
 const CONTEXT_GET_PUBLIC_VIS = 'custom_storage_public_visualizations_load';
 const CONTEXT_GET_PRIVATE_VIS = 'custom_storage_private_visualizations_load';
 const CONTEXT_GET_VIS = 'custom_storage_visualization_load';
--- a/packages/custom-storage/src/CustomStorage.ts
+++ b/packages/custom-storage/src/CustomStorage.ts
@@ -57,7 +57,8 @@ export class CustomStorage implements StorageRepository {
       COMMIT;
     `);
 
-    await Promise.all([this._publicSQLStorage.init(), this._privateSQLStorage.init()]);
+    const inits = await Promise.all([this._publicSQLStorage.init(), this._privateSQLStorage.init()]);
+    return inits[0] || inits[1];
   }
 
   public getVisualizations(): Promise<StoredVisualization[]> {
--- a/packages/custom-storage/src/CustomStorage.ts
+++ b/packages/custom-storage/src/CustomStorage.ts
@@ -63,7 +63,7 @@ export class CustomStorage implements StorageRepository {
     `);
 
     const event = new MetricsEvent(this._namespace, CONTEXT_INIT);
-    const inits = await Promise.all([this._publicSQLStorage.init(event), this._privateSQLStorage.init(event)]);
+    const inits = await Promise.all([this._publicSQLStorage.init({ event }), this._privateSQLStorage.init({ event })]);
 
     const storageHasBeenInitialized = inits[0] || inits[1];
     return storageHasBeenInitialized;
@@ -121,13 +121,13 @@ export class CustomStorage implements StorageRepository {
   public createVisualization(
     vis: Visualization,
     datasets: Array<Dataset | string>,
-    overwrite: boolean): Promise<StoredVisualization | null> {
+    overwriteDatasets: boolean): Promise<StoredVisualization | null> {
     this._checkReady();
 
     const target = vis.isPrivate ? this._privateSQLStorage : this._publicSQLStorage;
 
     const event = new MetricsEvent(this._namespace, CONTEXT_CREATE_VIS);
-    return target.createVisualization(vis, datasets, overwrite, event);
+    return target.createVisualization(vis, datasets, { overwriteDatasets, event });
   }
 
   public updateVisualization(vis: StoredVisualization, datasets: Dataset[]): Promise<StoredVisualization | null> {
@@ -214,7 +214,7 @@ export class CustomStorage implements StorageRepository {
   }
 
   private async _uploadDataset(dataset: Dataset, storage: SQLStorage, isPublic: boolean, overwrite: boolean) {
-    const storedDataset = await storage.uploadDataset(dataset, overwrite);
+    const storedDataset = await storage.uploadDataset(dataset, { overwrite });
 
     if (isPublic) {
       await storage.shareDataset(storedDataset.tablename);
--- a/packages/custom-storage/src/CustomStorage.ts
+++ b/packages/custom-storage/src/CustomStorage.ts
@@ -11,6 +11,7 @@ import {
 } from './StorageRepository';
 
 const CONTEXT_INIT = 'custom-storage-init';
+const CONTEXT_CREATE_VIS = 'custom-storage-create-visualization';
 
 export class CustomStorage implements StorageRepository {
   public static version: number = 0;
@@ -67,8 +68,6 @@ export class CustomStorage implements StorageRepository {
     const storageHasBeenInitialized = inits[0] || inits[1];
     return storageHasBeenInitialized;
   }
-
-
 
   public getVisualizations(): Promise<StoredVisualization[]> {
     this._checkReady();
@@ -127,7 +126,8 @@ export class CustomStorage implements StorageRepository {
 
     const target = vis.isPrivate ? this._privateSQLStorage : this._publicSQLStorage;
 
-    return target.createVisualization(vis, datasets, overwrite);
+    const event = new MetricsEvent(this._namespace, CONTEXT_CREATE_VIS);
+    return target.createVisualization(vis, datasets, overwrite, event);
   }
 
   public updateVisualization(vis: StoredVisualization, datasets: Dataset[]): Promise<StoredVisualization | null> {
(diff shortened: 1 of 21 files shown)


VictorVelarde:
This is still WIP, but I would love to have some feedback on the approach, @neokore  and @jesusbotella

packages/custom-storage/src/CustomStorage.ts
```diff
@@ -150,11 +160,11 @@ export class CustomStorage implements StorageRepository {
   }
 
   public uploadPublicDataset(dataset: Dataset, overwrite: boolean = false) {
-    return this.uploadDataset(dataset, this._publicSQLStorage, true, overwrite);
+    return this._uploadDataset(dataset, this._publicSQLStorage, true, overwrite);
```
neokore: Are we going to send metrics headers when uploading datasets?

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -165,51 +174,52 @@ export class SQLStorage {
...
+    const result: any = await this._sql.create(tableName, dataset.columns, { ifNotExists: false }, event);
 
     if (result.error) {
       throw new Error(`Failed to create table for dataset ${dataset.name}: ${result.error}`);
     }
 
     let copyResult: any;
     try {
-      copyResult = await this._sql.copyFrom(dataset.file, tableName, dataset.columns.map((column) => {
+      const fields = dataset.columns.map((column) => {
```
neokore: ⭐

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -390,7 +401,8 @@ export class SQLStorage {
     visId: string,
     datasets: Array<Dataset|string>,
     overwriteDatasets: boolean,
-    isPrivateVis: boolean) {
+    isPrivateVis: boolean,
```
neokore: Maybe is the moment to change some functions params to:
`functionName(mandatoryParam1, mandatoryParam2, {opts})`
to be able to add new params easier.
In this case:
`uploadAndLinkDatasetsTo(visId, datasets, {overwriteDatasets, isPrivateVis, event})`

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -455,37 +467,54 @@ export class SQLStorage {
...
     if (insertResult.error) {
       throw new Error('Failed to link dataset id to vis id');
     }
   }
 
-  private async _checkTable() {
+  /**
+   * Checks if all the tables for SQLStorage exist
+   */
+  private async _checkMissingTables() {
```
neokore: ❤️

packages/custom-storage/src/CustomStorage.ts
VictorVelarde: We have already added a MetricsEvent to 'init' and 'createVisualization' (which includes uploading and linking its datasets) calls.

The idea here is to add headers for the most relevant actions. Is `uploadPublicDataset` directly used in Kepler? 

In fact, it would probably be pretty relevant to list here, **the use of all custom-storage public methods from the client**. The methods being called would be probably quite less than the ones available at the public interface of the class

packages/custom-storage/src/sql/SQLStorage.ts
VictorVelarde: I thought about it, and yes, I completely agree with you on that refactor. It will allow us also to avoid having to fill some other 'optional' methods

packages/custom-storage/src/CustomStorage.ts
neokore: The current methods used in Kepler.gl are:
- `createVisualization` and `updateVisualization` from `CustomStorage` for saving maps
- `getVisualizations` and `getVisualization` from `CustomStorage` for loading maps
- `getVisualization` from `PublicStorageReader` that uses an own `SQL` class instance to load public maps

VictorVelarde: Then those are the main requests I will add events on

packages/custom-storage/src/sql/SQLStorage.ts
VictorVelarde: I have just added all events as optionals params inside generic Options objects, in case you want to have a look at it

packages/sql/src/CopyFromManager.ts
```diff
@@ -18,7 +18,7 @@ export class CopyFromManager extends RequestManager {
       body: file,
     };
 
-    const customHeaders = event ? event.getHeaders() : [];
+    const customHeaders = options.event ? options.event.getHeaders() : [];
```
neokore: I think we could use the null coalescing operator, At least I've used it before somewhere in this project.

VictorVelarde: TS 3.7 for the win!

VictorVelarde:
Ready for review. If it looks fine to you, we can merge it for a new release on Monday

packages/custom-storage/src/CustomStorage.ts
```diff
@@ -10,31 +10,36 @@ import {
   Visualization
 } from './StorageRepository';
 
+const CONTEXT_INIT = 'custom-storage-init';
```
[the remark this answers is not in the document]
VictorVelarde: I'm ok with adapting it to any standard we may follow. I have just used 'class' + 'method' names as a rule of thumb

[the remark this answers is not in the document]
esloho: The standard we're using is to use: ' subject' + 'action': `map_created`, `quota_exceeded`. So in this case if the action is the initialization of the custom storage then it should be: `custom_storage_init` or `toolkit_storage_init`

Also, notice the underscores :)

packages/custom-storage/src/CustomStorage.ts
```diff
@@ -10,31 +10,36 @@ import {
...
+const CONTEXT_CREATE_VIS = 'custom-storage-create-visualization';
```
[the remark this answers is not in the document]
esloho: By the same standard mentioned before, it should be one of these:
- `custom_storage_visualization_created`
- `toolkit_storage_visualization_created`

packages/custom-storage/src/CustomStorage.ts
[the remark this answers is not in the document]
esloho: Also, seeing the list of events in the PR description, the "get" ones could be renamed to use as action: `loaded` or `returned` ? Not sure the best option in that case, as you prefer really :D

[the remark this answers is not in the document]
VictorVelarde: I'm ok with following that standard: "subject + action" + underscores.

packages/custom-storage/src/CustomStorage.ts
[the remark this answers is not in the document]
VictorVelarde: @esloho @ilbambino Just a small nuance / comment: these headers are sent to the different APIs, for each of the "n-steps" that might be required to get a 'business-relevant' action. IMO they are not, strictly speaking, "the event", but the steps of it. 

For example, to create a visualization several calls will happen to the SQL API, for some checks and inserts in metadata tables + a call for 'copyFrom' for each of its datasets. So I'd rather "custom_storage_visualization_create" (the intention), to "custom_storage_visualization_visualization_created" (the fact, that would be completed after all the steps).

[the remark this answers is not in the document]
esloho: So if I understand correctly we should only register as event the one that marks as the task "completed", or do we need each "n-step" ?.

In the rest of projects the event marks the completion of some action, thus the past tense in the name. If this is not the case it makes sense to change it, as you decide :)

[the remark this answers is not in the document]
ilbambino: sounds good to me to use what @VictorVelarde suggests. If we implement the return code in the middle ware we could then have the full pic in analytics ;)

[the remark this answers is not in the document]
VictorVelarde: OK, I'm gonna rename the actions and also headers, following what we have talked about (see **https://github.com/CartoDB/toolkit/pull/56#issue-370074472** at the 🔝 for final names)

VictorVelarde:
This now should be ready to merge, once it is accepted on the Backend side.

Issue #23, opened 2019-10-08
Move to Typescript 3.7


rjimenezda:
Right now is in beta, so we should wait until it goes stable, but we could really use assert functions.

VictorVelarde:
We're on latest now. 👋

PR #58, opened 2020-02-04
Set custom number of retries and improve error messages for limits exceeded

neokore:


packages/auth/src/OAuth.ts
```diff
@@ -107,7 +107,8 @@ class OAuth {
       return null;
     }
 
-    return new Credentials('roman-carto', this._carto.accessToken.raw);
+    // TODO: Fix or remove!!!
+    return new Credentials(this._carto.userInfo, this._carto.accessToken.raw);
```
VictorVelarde: Wow, we had a strange const there hardcoded 😄 

Is the comment still relevant. I mean, userInfo is enough to fix it?

packages/custom-storage/src/CustomStorage.ts
```diff
@@ -22,9 +22,10 @@ export class CustomStorage implements StorageRepository {
     tableName: string,
     username: string,
     apiKey: string,
-    server: string = DEFAULT_SERVER) {
+    server: string = DEFAULT_SERVER,
+    maxRetries: number = DEFAULT_MAX_RETRIES) {
 
-    this._sqlClient = new SQL(username, apiKey, server);
+    this._sqlClient = new SQL(username, apiKey, server, {maxRetries});
```
VictorVelarde: Nice, an options param can work well here

packages/custom-storage/src/constants.ts
```diff
@@ -1,2 +1,3 @@
 export const DEFAULT_SERVER = 'https://{user}.carto.com/';
 export const DEFAULT_API_KEY = 'default_public';
+export const DEFAULT_MAX_RETRIES = 0;
```
VictorVelarde: I have just removed this file in #57 (extracting the first 2 constants to the new core package). We can keep this file as it is, with just the new const or maybe extract it (to the CustomStorage.ts??)

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -390,41 +394,41 @@ export class SQLStorage {
     overwriteDatasets: boolean,
     isPrivateVis: boolean) {
 
-    for (const dataset of datasets) {
-      let tableName: string;
+      for (const dataset of datasets) {
```
VictorVelarde: Not sure why we have these code indentation changes. IMO we should automate this on committing, in order to avoid this kind of formatting issues

packages/sql/src/RequestManager.ts
```diff
@@ -13,7 +13,7 @@ interface FetchArgs {
 
 const UNKNOWN = -1;
 const NO_RETRY = -1;
-const MAX_RETRIES = 5;
+const MAX_RETRIES = 0;
```
VictorVelarde: maybe call this DEFAULT_MAX_RETRIES?

packages/sql/src/RequestManager.ts
```diff
@@ -159,7 +166,8 @@ export class RequestManager {
         if (!data.error) {
           resolve(data);
         } else {
-          reject(data.error);
+          const message = data && data.error && data.error.length ? data.error[0] : 'Unknown error';
```
VictorVelarde: Hopefully some chaining soon, if we upgrade Typescript #23

packages/sql/src/RequestManager.ts
```diff
@@ -26,14 +26,17 @@ export class RequestManager {
...
-  constructor(credentials: Credentials) {
+  constructor(credentials: Credentials, {maxRetries}: {maxRetries?: number} = {}) {
     const { username, apiKey, server } = credentials;
 
     this._username = username;
     this._apiKey = apiKey;
     this._server = server.replace('{user}', username);
     this._queue = [];
+
+    if (maxRetries || maxRetries === 0) { this._maxRetries = maxRetries; }
```
VictorVelarde: I don't get the `(maxRetries === 0)` comparison. Could you explain it a bit?

packages/toolkit/src/App.ts
```diff
@@ -2,15 +2,18 @@ import { CustomStorage, PublicStorageReader } from '@carto/toolkit-custom-storag
 import { SQL } from '@carto/toolkit-sql/dist/types/Client';
 const DEFAULT_SERVER = 'https://{user}.carto.com/';
 const DEFAULT_NAMESPACE = 'toolkit';
+const DEFAULT_MAX_RETRIES = 0;
 
 export interface AppOptions {
   namespace: string;
   server: string;
+  maxRetries: number;
```
VictorVelarde: a bit more explicit, like `maxApiRequestsRetries`?

VictorVelarde:
Don't forget to lint:fix !

packages/sql/src/RequestManager.ts
neokore: To let you set it to 0 in case we define another default value. The falsy value of `maxRetries` would not pass the test. Anyway, I've change it to a restricted test.

packages/toolkit/src/App.ts
neokore: Sure!

packages/sql/src/RequestManager.ts
neokore: Changed all to `DEFAULT_MAX_API_REQUESTS_RETRIES` and all files get it from SQL module

packages/sql/src/RequestManager.ts
neokore: I forgot, yes. But I think it works. Let's try

VictorVelarde: Not a priority though, it just came to my mind 😄

VictorVelarde:
LGTM, feel free to merge afte making our linter happy again

Issue #59, opened 2020-02-10
Create new Viz package


VictorVelarde:
A package to hold classes and utilities to build maps with deck-gl & carto

VictorVelarde:
A first version has been created in web-sdk branch

PR #60, opened 2020-02-11
Create a new viz package [WIP]

VictorVelarde:
Deals with #59 

New module to host the CartoTileLayer wrapper, whose mission is to instantiate a deck.gl layer from CARTO.

ch: Layers [ch57706]


packages/viz/src/CartoTileLayer.ts
```diff
@@ -1,88 +1,84 @@
...
 
-    this._source = source;
+    this._credentials = credentials;
+    this._mapsClientInstance = new Maps(this._credentials);
+    this._mapSource = new Source(source);
 
-    const { deckInstance, mapOptions, credentials } = options;
+    // Map Instantiation
+    this._mapOptions = Object.assign({}, defaultMapOptions, mapOptions);
+    this._mapInstantiation = this._mapsClientInstance.instantiateMapFrom(
```
VictorVelarde: It's fine to have the map instantiation here 👍 
One relevant thing to consider is (if / when) we will need to 're-instantiate' a map. For example, will we have a 'setSource' to allow updating the Layer?

packages/viz/src/CartoTileLayer.ts
```diff
@@ -1,88 +1,84 @@
...
+      pointRadiusMinPixels: 5,
+      urlTemplates,
+      uniquePropertyName: 'cartodb_id',
+      renderSubLayers: (props: any) => new deckSublayer({ ...props })
     });
-    this._credentials = credentials || Credentials.default;
 
-    // this._style = style;
-    this._mapsClient = new Maps(this._credentials);
+    return new MVTTileLayer(layerProperties);
```
VictorVelarde: Ok, I see we won't need any deck reference here, just the reference to our class. 👍

packages/viz/src/CartoTileLayer.ts
```diff
@@ -1,88 +1,84 @@
...
-      options = Object.assign({}, this._mapOptions, { dataset: this._source });
-    } else {
-      options = Object.assign({}, this._mapOptions, { sql: this._source });
-    }
+function buildInstantiationOptions(
+  { mapOptions, mapSource }: { mapOptions: MapOptions, mapSource: Source}
+) {
+  return {
+    ...mapOptions,
+    ...mapSource.getSourceOptions()
```
VictorVelarde: I see this more like an implementation detail of the Maps API. Should we consider moving these methods to the maps package itself?

packages/viz/src/CartoTileLayer.ts
```diff
@@ -1,88 +1,84 @@
...
+}
 
-    return await this._mapsClient.instantiateMapFrom(options);
-  }
+interface LayerOptions {
+  credentials?: Credentials;
+  mapOptions?: MapOptions;
 }
 
+interface MapOptions {
```
VictorVelarde: Same here, shouldn't we move this interface to 'Maps' package?

VictorVelarde: Maybe we could have a method to build a 'default' object with 2048 values here, but keep the type in the 'maps' package.

packages/viz/src/Source.ts
```diff
@@ -0,0 +1,25 @@
+class Source {
```
VictorVelarde: Yeah, I also think this Source will be useful as a separated class.

One thing that comes to my mind is that it might even have 'data-related' methods, by using internally an SQL client reference ('sql' api package). Thinking about 'getGeometryType', 'getBounds', 'getStats'....

packages/viz/src/CartoTileLayer.ts
jesusbotella: I guess so! I left them here because they were here before, but I agree that it makes more sense to move them there.

packages/viz/src/CartoTileLayer.ts
jesusbotella: Yes, we can export it from our Maps package.

packages/viz/src/Source.ts
jesusbotella: That might make sense! We can do that

packages/viz/src/CartoTileLayer.ts
jesusbotella: I was rethinking this, and I'm not sure about this. If we move this function to Maps package, we should move our Source declaration there as well, so that the Maps package is aware of types.

VictorVelarde:
I think we should have master --> web-sdk --> feature branches. So any new (smaller) PR should point to 'web-sdk', instead of master

IMO web-sdk should contain, by the end of this iteration:
- new viz module
- new viz features (PRs for styles)
- webpack as the only bundler, for all packages (maybe another PR)?

jesusbotella:
Aren't we going to merge all of the new features to master? We are going to end up having a PR with lots of lines changed. I vote for merging this PR to master, and continue from there.

Also, when did we start using yarn? I noticed that there's a `yarn.lock` file that was created and I updated, but my change was by mistake probably.

jesusbotella:
@VictorVelarde Could you please give this another 👀?
I'd like to merge it, and then merge the styles PR, so that we don't keep piling things up

packages/viz/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@carto/toolkit-viz",
+  "version": "0.0.1-rc.13",
```
VictorVelarde: I think we should align with current version, thus **0.0.1-rc.16**

VictorVelarde: Once we rebase master, we should update this to the current **0.0.1-rc.16**, also its dependencies, and run the tests****

packages/viz/src/Layer.ts
```diff
@@ -0,0 +1,90 @@
...
+      getLineColor: [192, 0, 0],
+      getFillColor: [200, 120, 80],
+      lineWidthMinPixels: 1,
+      pointRadiusMinPixels: 5,
+      urlTemplates,
+      uniquePropertyName: 'cartodb_id'
+    };
+
+    const layerProperties = Object.assign({}, defaultLayerProps, overridenStyleProps);
+    return new MVTTileLayer(layerProperties);
```
VictorVelarde: ❤️ I love this line... 😄

VictorVelarde:
Overall it looks ✨pretty good to me, and it seems like it doesn't collide with current work related to customStorage & kepler.gl, so let's go ahead. As we know, we have to apply more steps (refactor, webpack use, examples...), but let's do that step by step. 

I would advise rebasing current master, and asking @neokore for a quick look / aprovement before merging into master.

PR #61, opened 2020-03-02
Escape chars for string fields and check initialization for storage

neokore:


VictorVelarde:
LGTM. Go ahead with it if you're in a hurry, but please add a task to this week for some unit tests

PR #63, opened 2020-03-04
Fix initialization checks

neokore:


packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -95,7 +95,9 @@ export class SQLStorage {
    * Checks if storage tables are created
    */
   public async isInitialized() {
-    return !this._checkMissingTables();
+    const hasMissingTables = await this._checkMissingTables();
+    this._isReady = !hasMissingTables;
+    return !hasMissingTables;
```
VictorVelarde:
```suggestion
    return this._isReady;
```

packages/custom-storage/src/CustomStorage.ts
```diff
@@ -212,8 +212,10 @@ export class CustomStorage implements StorageRepository {
     await this._publicSQLStorage.destroy();
   }
 
-  private isInitialized() {
-    return this._publicSQLStorage.isInitialized();
+  public async isInitialized() {
+    const inits = await Promise.all([this._publicSQLStorage.isInitialized(), this._privateSQLStorage.isInitialized()]);
```
VictorVelarde: Why do we need to check both storages?

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -95,7 +95,9 @@ export class SQLStorage {
    * Checks if storage tables are created
    */
   public async isInitialized() {
-    return !this._checkMissingTables();
+    const hasMissingTables = await this._checkMissingTables();
+    this._isReady = !hasMissingTables;
```
VictorVelarde: ok, I see, this._isReady was not yet correctly setup...

packages/custom-storage/src/sql/SQLStorage.ts
```diff
@@ -532,16 +534,23 @@ export class SQLStorage {
...
+        const checksTablesAreReady = requiredTables.map((table) => this._sql.query(`SELECT to_regclass('${table}')`));
+        const results = await Promise.all(checksTablesAreReady);
+        const missingTables = results.some((response: any) => {
+          const tableIsMissing = (response.rows[0].to_regclass === null);
+          return tableIsMissing;
+        });
+
+        resolve(missingTables);
+      } catch (err) {
+        reject(err);
```
VictorVelarde: Out of curiosity, what kind of errors have you found here?

neokore: No errors yet, but as long as I have to create a Promise, I always prefer to make use of the reject function for possible errors

packages/custom-storage/src/CustomStorage.ts
neokore: In order to set both as ready

PR #64, opened 2020-03-09
Add default and static styles

jesusbotella:
Add default styles like the ones in CARTOframes and static styles.


packages/viz/src/lib/Layer.ts
```diff
@@ -42,32 +50,31 @@ export class Layer {
...
+    const urlData = metadata.url.vector;
     const urlTemplates = urlData.subdomains.map(
       (subdomain: string) => urlData.urlTemplate.replace('{s}', subdomain)
     );
 
-    return { urlTemplates };
+    const geometryType = metadata.layers[0].meta.stats.geometryType.split('ST_')[1];
+
+    return { urlTemplates, geometryType };
   }
```
jesusbotella: Should we abstract all this parsing and map instantiation to another class/whatsoever within our visualization module?

VictorVelarde: I would say this corresponds to maps package, who knows the internal about map instantiation and the metadata fields

packages/viz/src/lib/errors/carto-error.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @typedef {Object} CartoError
+ * @property {String} message - A short error description
+ * @property {String} name - The name of the error "CartoError"
+ * @property {String} type - The type of the error "CartoError"
+ * @property {Object} originalError - An object containing the internal/original error
+ *
+ * @event CartoError
+ * @api
+ */
+export default class CartoError extends Error {
```
VictorVelarde: What do you think about moving CartoError to core?

packages/viz/src/lib/style/helpers/ramps.ts
```diff
@@ -0,0 +1,34 @@
+import {CartoStylingError, stylingErrorTypes} from '../../errors/styling-error';
+
+export function numericRamp(featureName: string, numericValues = [], colors = []) {
+  if (!featureName) {
+    throw new CartoStylingError('Feature name is missing', stylingErrorTypes.PROPERTY_MISSING);
+  }
+
+  const lengthMismatch = numericValues.length !== colors.length;
```
VictorVelarde: Depending on open / closed intervals, we might want to compare to colors +/- 1

I would start with closed intervals on right sides, so 0-100-500 would be useful to define 2 buckets (2 colors) like:
- (a) 0 to 100 (>=0 to <100)
- (b) 100 to 500 (>=100 to <500)

So 0 will be included in (a), 100 in (b) and 500 outside of intervals. To assign 500 a color, a new interval would be needed (eg 0, 100, 500, 1000).

How do you see this approach?

packages/viz/examples/index.html
```diff
@@ -23,38 +23,33 @@
...
-    <script src="https://unpkg.com/deck.gl@8.0.14/dist.min.js"></script>
+    <script src="https://unpkg.com/deck.gl@8.1.0-alpha.6/dist.min.js"></script>
 
     <!-- CARTO Deck.gl's Tile Layer -->
     <script src="/dist/viz.umd.js"></script>
 
     <script>
       async function initialize() {
-        cartoToolkitViz.setDefaultCredentials()
+        carto.setDefaultCredentials({ username: 'jbotella' });
```
VictorVelarde: let's try to use a generic username for examples. We can reuse 'public', like for the blog, or maybe create a new 'web-sdk' user

packages/maps/src/Client.ts
```diff
@@ -92,3 +92,56 @@ export interface MapOptions {
     geometryType: boolean
   };
 }
+
+export interface MapInstance {
```
VictorVelarde: Really nice addition. I thought about integrating it at the beginning, but I didn't.

packages/maps/src/Client.ts
```diff
@@ -92,3 +92,56 @@ export interface MapOptions {
...
+  layergroupid: string;
+  last_updated: string;
```
VictorVelarde: last_updated ? vs updated_at --> Is the spec not updated? https://carto.com/developers/maps-api/reference/#operation/instantiateAnonymousMap

packages/maps/src/Client.ts
```diff
@@ -17,22 +17,22 @@ export class Maps {
...
           sql: sql || `select * from ${dataset}`,
-          vector_extent: vector_extent || 2048,
-          vector_simplify_extent: vector_simplify_extent || 2048
+          vector_extent,
+          vector_simplify_extent,
+          metadata
         }
-      }],
-      version: '1.3.1'
+      }]
```
VictorVelarde: Are you sure we don't need 'version' field anymore on mapConfig? What's the behavior without it / latest?

packages/viz/__tests__/Layer.spec.ts
```diff
@@ -46,7 +53,7 @@ describe('Layer', () => {
         'https://notdefaultserver.com'
       );
 
-      const layer = new Layer(DEFAULT_DATASET, { credentials });
+      const layer = new Layer(DEFAULT_DATASET, {}, { credentials });
```
VictorVelarde: I guess this 2nd param is Style. Should we consider adding it to the 'options' bag?

packages/viz/examples/styles/default_styles/points.html
```diff
@@ -0,0 +1,58 @@
...
+    <!-- Map libraries-->
+    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.0.0/mapbox-gl.js"></script>
+    <script src="https://unpkg.com/deck.gl@8.1.0-alpha.6/dist.min.js"></script>
+
+    <!-- CARTO Deck.gl's Tile Layer -->
+    <script src="/dist/viz.umd.js"></script>
+
+    <script>
+      async function initialize() {
+        carto.setDefaultCredentials({ username: 'cartojs-test' });
```
VictorVelarde: We should try to homogenize users data into just one, selected user

packages/viz/src/lib/Layer.ts
```diff
@@ -1,38 +1,45 @@
...
 
   private _layerSource: Source;
+  private _layerStyles: Style;
   private _layerOptions: MapOptions;
-  private _layerInstantiation: Promise<any>; // TODO: Change to a proper definition
+  private _layerInstantiation: Promise<MapInstance>;
 
-  constructor(source: string, options: LayerOptions = {}) {
-    const { mapOptions, credentials = defaultCredentials } = options;
+  constructor(source: string, styles = {}, options: LayerOptions = {}) {
```
VictorVelarde: how about moving styles to options?

packages/viz/src/lib/errors/styling-error.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export class CartoStylingError extends CartoError {
+  constructor(message: string, type = stylingErrorTypes.DEFAULT) {
+    super({ message, type });
+    this.name = 'CartoStylingError';
+  }
+}
+
+export const stylingErrorTypes = {
+  DEFAULT: '[Error]',
```
VictorVelarde: This is the approach we had for carto-vl, and I added a lot of 'types'... do you feel like is it too complex?

packages/viz/src/lib/style/defaultStyles.ts
```diff
@@ -0,0 +1,42 @@
+import {Style} from './Style';
```
VictorVelarde: Nice defaults!

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -0,0 +1,35 @@
+import { getColors, validateParameters } from './utils';
+
+export function colorBinsStyle(
+  featureName: string,
+  { bins = defaultOptions.bins, binColors = defaultOptions.binColors }: ColorBinStyleOptions = defaultOptions
```
VictorVelarde: Nit, should we change ColorBinStyleOptions to ColorBinsStyleOptions with 's'?

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -0,0 +1,35 @@
...
+      (featureValue >= definedValue) && (featureValue < valuesArray[currentIndex + 1]);
+
+    const featureValueIndex = ranges.findIndex(rangeComparison);
+    return rgbaColors[featureValueIndex];
+  };
+
+  return { getFillColor };
+}
+
+interface ColorBinStyleOptions {
```
VictorVelarde: 🔝

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -0,0 +1,35 @@
...
+  const ranges = [...bins, Number.MAX_SAFE_INTEGER];
+  const rgbaColors = getColors(binColors, bins.length);
+
+  const getFillColor = (feature: Record<string, any>) => {
+    const featureValue: number = feature.properties[featureName];
+
+    // If we want to add various comparisons (<, >, <=, <=) like in TurboCARTO
+    // we can change comparison within the arrow function to a comparison fn
+    const rangeComparison = (definedValue: number, currentIndex: number, valuesArray: number[]) =>
+      (featureValue >= definedValue) && (featureValue < valuesArray[currentIndex + 1]);
```
VictorVelarde: >= and <, I like it. We should probably document this behaviour in the function, as it is probably one of the most common confusions when dealing with bins

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    const featureValueIndex = ranges.findIndex(rangeComparison);
+    return rgbaColors[featureValueIndex];
+  };
+
+  return { getFillColor };
```
VictorVelarde: This colorBinsStyle is, as it is, useful for fillColor, which is probably the most common request. Maybe we could make it a bit more flexible in the future (not now), so the user can choose which 'color' property he will get ('getFillColor' by default vs 'getColor')... but I would wait a bit before introducing more complexity

packages/viz/src/lib/style/helpers/color-categories-style.ts
```diff
@@ -0,0 +1,33 @@
...
+    categoryColors = defaultOptions.categoryColors
+  }: ColorCategoriesStyleOptions = defaultOptions
+) {
+  validateParameters(featureName, categories, categoryColors);
+
+  const rgbaColors = getColors(categoryColors, categories.length);
+  const categoriesWithColors = convertArrayToObjectWithValues(categories, rgbaColors);
+
+  const getFillColor = (feature: Record<string, any>) => {
+    // TODO: Add color for other categories
```
VictorVelarde: and 'null' data (a different PR in the future, with improvements...)

packages/viz/src/lib/style/helpers/color-categories-style.ts
```diff
@@ -0,0 +1,33 @@
...
+    const category = feature.properties[featureName];
+    return categoriesWithColors[category];
+  };
+
+  return { getFillColor };
+}
+
+interface ColorCategoriesStyleOptions {
+  categories: string[];
```
VictorVelarde: This is where we can add in the future another type, defining a method to get the categories, like 'top' + number '5' or similar

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { CartoStylingError, stylingErrorTypes } from '../../errors/styling-error';
+import { hexToRgb } from '../../utils/colors';
+import { getColorPalette } from '../palettes';
+
+export function validateParameters(featureName: string, values: number[] | string[], colors: string[] | string) {
+  if (!featureName) {
```
VictorVelarde: how about 'featureProperty' instead of 'featureName'?

packages/viz/src/lib/style/palettes/palette.ts
```diff
@@ -0,0 +1,81 @@
+import { SubPalette } from 'cartocolor';
+
+/**
+ * Color palettes.
```
VictorVelarde: Nice reuse here!

VictorVelarde:
I think that this, the first feature addition to web-sdk, should also start the CHANGELOG.

Let's start with something manual, 1 file for the whole proj, and prepending the package name, like in airship. Something like this?

## Development
### Features
- [viz] Default style for layers
- [viz] colorBinsStyle and colorCategoriesStyle with manual mode
### Bug fixes
-
### BREAKING CHANGES
-

packages/viz/src/lib/style/helpers/ramps.ts
jesusbotella: I thought of doing that at first because it seemed more intuitive for me, but programming it the other way was easier and TurboCARTO did it with open intervals.

But I think it's better with closed intervals.

jesusbotella:
Sorry about this, I wanted to push some commits before but I couldn't because of GitHub downtime. I updated helpers to have others' color, null color, and closed intervals.

Sorry for making you to review it again :(

packages/maps/src/Client.ts
jesusbotella: `last_updated` is the one that I took from an actual request. Seems like that the spec not updated :(

packages/maps/src/Client.ts
jesusbotella: It is `1.0.0` as per docs. I removed it because I saw that CARTO-VL didn't include that in the request. But thinking it right now, I think we should keep it just for safety.

packages/viz/__tests__/Layer.spec.ts
jesusbotella: Yes, it is our style definition. But I think we should not add it to our options object, just for simplicity, it's much better to specify it like:
```js
const layer = new carto.viz.Layer(DEFAULT_DATASET, carto.viz.colorBinsStyle('column'))
```
rather than:
```js
const layer = new carto.viz.Layer(DEFAULT_DATASET, {
  style: carto.viz.colorBinsStyle('column')
})
```

What we can do is to detect which parameter comes second and then change our behaviour. I'm gonna write it down in Clubhouse.

packages/viz/src/lib/Layer.ts
jesusbotella: I replied this in a response above.

packages/viz/src/lib/errors/styling-error.ts
jesusbotella: I think it's good so far. We need to keep it simple, but an enumeration with the current error types is nice.

packages/viz/src/lib/style/helpers/color-bins-style.ts
jesusbotella: Yes, sure! We need to be consistent!

packages/viz/src/lib/style/helpers/color-bins-style.ts
jesusbotella: Right now, it's static, it doesn't change. But what I wanted to note it's how it should change when adding that feature.

packages/viz/src/lib/style/helpers/color-bins-style.ts
jesusbotella: Yes, I have just added the most common style specification that all (or almost all) Deck.gl layers use, so that the developer can swap layers and get what they expect.

packages/viz/src/lib/style/helpers/color-categories-style.ts
jesusbotella: Added in this same PR!

packages/viz/src/lib/Layer.ts
jesusbotella: I am gonna write it down in Clubhouse as well, because this refactor will add a lot of changed lines in this PR.

packages/viz/__tests__/Layer.spec.ts
VictorVelarde: Mmm, I see. Well I won't bother right now with 'auto-detecting'. Let's leave it as it is.

packages/viz/src/lib/style/helpers/color-bins-style.ts
VictorVelarde: Yes, perfect. I was just saying adding a small description about closed intervals to the function header

packages/viz/src/lib/style/helpers/color-categories-style.ts
VictorVelarde: That's great!

jesusbotella:
I have just pushed all commits addressing review comments. I think we're good to go!

About the changelog, I think we should make it automatic like we did in Airship, following commits convention. It's much easier and we don't need to bother adding changes manually.

packages/viz/examples/styles/helpers/colorBinsStyle.html
```diff
@@ -0,0 +1,66 @@
...
+      }
+    </style>
+  </head>
+
+  <body>
+    <section id="map"></section>
+
+    <!-- Map libraries-->
+    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.0.0/mapbox-gl.js"></script>
+    <script src="https://unpkg.com/deck.gl@8.1.0-alpha.6/dist.min.js"></script>
```
VictorVelarde: It looks like we can now update and remove alpha, using official 8.1.0 !! 🎊

packages/viz/examples/styles/helpers/colorBinsStyle.html
```diff
@@ -0,0 +1,66 @@
...
+        height: 100vh;
+      }
+    </style>
+  </head>
+
+  <body>
+    <section id="map"></section>
+
+    <!-- Map libraries-->
+    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.0.0/mapbox-gl.js"></script>
```
VictorVelarde: Regarding to mapbox-gl, I have reviewed deck.gl repo and they are not very consistent with versions, but in latests examples they use `https://api.tiles.mapbox.com/mapbox-gl-js/v1.4.0/mapbox-gl.js`

As a rule of thumb, I guess we could always use the latest stable mapbox, unless deck.gl becomes more explicit. Currently, latests stable = 1.8.1

packages/viz/examples/index.html
```diff
@@ -23,38 +23,33 @@
...
-    <script src="https://unpkg.com/deck.gl@8.0.14/dist.min.js"></script>
+    <script src="https://unpkg.com/deck.gl@8.1.0-alpha.6/dist.min.js"></script>
 
     <!-- CARTO Deck.gl's Tile Layer -->
     <script src="/dist/viz.umd.js"></script>
 
     <script>
       async function initialize() {
-        cartoToolkitViz.setDefaultCredentials()
+        carto.setDefaultCredentials({ username: 'public' });
```
VictorVelarde: nice change!

packages/viz/examples/index.html
```diff
@@ -23,38 +23,33 @@
...
-            deckGLLayer
-          ]
+          controller: true
         });
 
-        window.deckInstance = deckInstance;
+        const distritosLayer = new carto.Layer('ne_50m_admin_0_countries');
+        distritosLayer.addTo(deckMap);
+
+        window.deckMap = deckMap;
```
VictorVelarde: Is this line required for the example?

packages/viz/examples/styles/helpers/colorBinsStyle.html
jesusbotella: I'm gonna update the examples. That version was there since the very beginning, so I'm gonna update them, and point our `@deck.gl/geo-layers` dependency to 8.1.0 to finally get the polygon clipping.

packages/viz/examples/styles/helpers/colorCategoriesStyle.html
```diff
@@ -0,0 +1,73 @@
...
+          },
+          controller: true,
+        });
+
+        const roomTypeStyle = carto.colorCategoriesStyle('room_type', {
+          categories: ['Entire home/apt', 'Private room', 'Shared room'],
+
+          // categoryColors accepts an array of HEX color strings,
+          // or a string pointing to a CARTOColors palette
+          categoryColors: ['#88ccee', '#cc6677', '#ddcc77'],
```
VictorVelarde: Should we just rename `categoryColors` to `colors` parameter? This array is just in the context of colorCategoriesStyle, so...

packages/viz/package.json
```diff
@@ -40,6 +40,7 @@
     "ts-loader": "^6.2.1",
     "typescript": "^3.6.3",
     "tslib": "^1.10.0",
+    "@danmarshall/deckgl-typings": "3.4.6",
```
VictorVelarde: Cool, I didn't know about them.

I have checked the proj, and it looks like they will try to get up to v8 (but it looks like they don't have many contributors...). Not sure if in the meantime using 'old' typings (for v 7.x.x.) could cause any issue.

packages/viz/src/declarations/deckgl.d.ts
```diff
@@ -0,0 +1,9 @@
+// TODO: Add proper module declaration;
```
VictorVelarde: In the future maybe we could contribute it to https://www.npmjs.com/package/@danmarshall/deckgl-typings

packages/viz/src/lib/style/helpers/color-bins-style.ts
VictorVelarde: Small detail: this filename uses color-bins-style.ts naming, should we go colorBinsStyle.ts ?

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const lengthMismatch = lengthComparisonFn();
+  const colorsIsNotString = typeof colors !== 'string';
+
+  if (colorsIsNotString && lengthMismatch) {
+    throw new CartoStylingError(
+      'Numeric values for ranges length and color length do not match',
+      stylingErrorTypes.PROPERTY_MISMATCH
+    );
+  }
```
VictorVelarde: Perhaps we can add another validation here, if colors === 'string' (palette name), that the palette exists in cartocolors., reusing code from 'getColors' method

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+      rgbaColors: rgbaColors.map(hexToRgb),
+      othersColor: othersColor ? hexToRgb(othersColor) : void(0)
+    };
+  }
+
+  return { rgbaColors: (colorProperty as string[]).map(hexToRgb) };
+}
+
+// Extracted from https://github.com/CartoDB/carto-vl/blob/develop/src/renderer/viz/expressions/utils.js#L53
+export function hexToRgb(hex: string) {
```
VictorVelarde: This is something that can be a potential (small) contribution to deckgl core

packages/maps/src/Client.ts
VictorVelarde: I created an issue at dev-center for them

packages/viz/package.json
jesusbotella: Not right now, because we're only using it for this declaration:

https://github.com/CartoDB/toolkit/blob/19d3b65498e0c381d90176d01f0c1c2fcd73ac6a/packages/viz/src/declarations/deckgl.d.ts#L3-L9

packages/viz/src/declarations/deckgl.d.ts
jesusbotella: Yes, I would have added our declarations to the repo, but as they were working on 8.0 declarations I didn't want to disturb

packages/viz/src/lib/style/helpers/utils.ts
jesusbotella: Honestly, I prefer to keep them separated, because here we are like validating the type/shape, rather than if they're valid or not.

If we were to add that kind of validation, we'd need to add some other ones like checking if the HEX string is valid, and I prefer not to put them here.

packages/viz/examples/styles/helpers/colorCategoriesStyle.html
jesusbotella: It's a good idea now that we have `nullColor` and `othersColor`. I'm gonna rename `binColors` and `categoryColors` to just colors.

packages/viz/src/lib/style/palettes/palette.ts
jesusbotella: Sadly, I have not been able to found any. I'm gonna create another branch for testing the code.

packages/viz/src/lib/style/helpers/color-bins-style.ts
jesusbotella: I mixed naming conventions across files 🤦‍♂ 

Many people follow `-` separation and all lowercase, and it's the one that I like the most. But I need to change `defaultStyles.ts` to `default-styles.ts`.

What do you think?

VictorVelarde: default-styles is better for reading, isn't it?

packages/viz/src/lib/style/palettes/palette.ts
VictorVelarde: don't stop now for the PoC, as palette is quite tested (on vl)

packages/viz/src/lib/style/helpers/utils.ts
VictorVelarde: fair point

VictorVelarde:
@jesusbotella This PR looks really great! I don't want to block it for merging to web-sdk, so we can get a cool PoC demo, so feel free to merge, as soon it's green
 
Any non basic change can wait a bit (as eg. adding the conventional commit changelog), and let's add a short story for them if required

packages/viz/package.json
```diff
@@ -47,7 +47,7 @@
     "@deck.gl/layers": "^8.1.0",
     "@deck.gl/geo-layers": "^8.1.0",
     "@deck.gl/mesh-layers": "^8.1.0",
-    "@luma.gl/core": "8.1.0-alpha.4",
```
VictorVelarde: 😄

jesusbotella:
Don't worry about that. I think I addressed all review comments and wrote down other ones in CH.

I'm gonna wait until https://github.com/CartoDB/toolkit/pull/68 review, because this branch has some tests fixed in the other PR 🙈. So if you could take a look at the other PR, it would be great.

Once the other PR is approved, I'll merge it into this, and then merge this into `web-sdk` branch.

PR #68, opened 2020-03-16
Change bundler to export ESM and add viz package to toolkit package

jesusbotella:
This PR modifies some important things such as bundling for toolkit and visualization packages.

When changing from `rollup` bundling to `webpack` (because of Deck.gl's compatibility with `rollup`), we lost the ability of exporting `esm` and `cjs` modules which are useful for developers that want to import our packages within their current build systems and tree-shake the modules.

So, I have added transpilation for `@carto/toolkit` and `@carto/toolkit-viz` to `esm` and `cjs` through Babel. This will not output a bundle (as rollup did) but separated files that bundlers are able to handle.

Another important change is that I removed `browser` property from all modules' `package.json`, because it was pointing to UMD bundles. As Webpack uses that property to get external modules, it's not OK to point to UMD, but to ESM bundle. So I removed it, and now webpack should use `module` field from package. We should leave UMD bundle for CDN only.

As a consequence, I had to modify rollup config to not take the proper file name from `package.json` but from another object. And I leveraged that opportunity to update rollup dependency version.

That said, everything should be transparent and we should not have to make any changes to the imports that we already have outside the library.


VictorVelarde:
How difficult would be to completely remove rollup in favour of **webpack** in this PR? 
I understand webpack is a must for viz, due to deck.gl compatibility, so I feel like we should have just one shared tool, the sooner the better... Thoughts?

jesusbotella:
I avoided doing it because I preferred using rollup, and I hoped that we could use it some day, but being rational it is true that we should remove it altogether.

VictorVelarde:
Yes, I aggree with you. It's up to you doing it here or as a next PR, but it will be simpler for all

bundle/webpack.base.js
```diff
@@ -0,0 +1,56 @@
...
+  // Depending on mode Webpack will apply different things
+  // on final bundle.
+  mode,
+
+  entry: './src/index.ts',
+
+  optimization: { usedExports: true },
+
+  externals: {
+    '@luma.gl/core': {
```
VictorVelarde: I see the `deck` case, but why do we need to set `luma` as externals? what's the use case?

packages/auth/package.json
```diff
@@ -17,12 +16,14 @@
     "lint": "tslint -c tslint.json {src,__tests__}/**/*.{ts,tsx}",
     "lint:fix": "tslint --fix -c tslint.json {src,__tests__}/**/*.{ts,tsx}",
     "prebuild": "rimraf dist/",
-    "build": "rollup -c rollup.config.ts",
-    "build:debug": "rollup -c rollup.config.ts --configDebug",
+    "build": "npm run build:esm && npm run build:cjs && npm run build:umd",
```
VictorVelarde: Build tasks look fine. But we're loosing `build:debug`, right?

packages/viz/examples/styles/default_styles/lines.html
```diff
@@ -29,11 +29,11 @@
     <script src="https://unpkg.com/deck.gl@8.1.0/dist.min.js"></script>
 
     <!-- CARTO Deck.gl's Tile Layer -->
-    <script src="/dist/viz.umd.js"></script>
+    <script src="/dist/umd/index.min.js"></script>
 
     <script>
       async function initialize() {
-        carto.setDefaultCredentials({ username: 'public' });
+        cartoViz.setDefaultCredentials({ username: 'public' });
```
VictorVelarde: we'll move this in the future to upper folder for examples, and then we'll use `carto.setDefaultCredentials` + `carto.viz.Layer` ...

packages/viz/package.json
```diff
@@ -1,28 +1,33 @@
...
-  "browser": "dist/viz.umd.js",
-  "main": "dist/viz.cjs.js",
-  "module": "dist/viz.esm.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
   "directories": {
     "lib": "src",
-    "test": "__tests__"
+    "test": "__tests__",
+    "example": "examples"
```
VictorVelarde: Out of curiosity... what's the purpose of directories.example?

packages/viz/src/index.ts
```diff
@@ -1,3 +1,5 @@
+// TODO: Should we export setDefaultCredentials from here,
```
VictorVelarde: IMO just from top package

bundle/webpack.base.js
jesusbotella: In fact, I am gonna add all `@deck.gl/**` modules to external dependencies as we are loading deck.gl via HTML script.

`luma.gl` is bundled within `@deck.gl/core`, so there's no need of including it twice, and it throws if you have two different versions imported.

packages/auth/package.json
jesusbotella: I don't think we were using it that much, right? Webpack doesn't have that debug mode by default as far as I know.

We have two modes now:
- By default, development mode is active.
- When using `NODE_ENV=production npm run build`, it will bundle our scripts for production. That's what we should put in our release script when we have it.

packages/viz/package.json
jesusbotella: I reviewed package.json specification and it had that property as well, so I decided to set it to our examples folder. It's not used anywhere (said in docs) but given that we had those two others I decided to fill it in.

jesusbotella:
We don't have any package-lock.json files within packages. We just have one in our root folder that is the one updated within this PR.

I changed the review suggestions and answered questions!

VictorVelarde:
Maybe its presence was a problem in my local env... I tested this on top of the style PR

PR #72, opened 2020-03-23
Case sensitive field names in sql queries #47

neokore:


jesusbotella:
Just a silly question, but why are we adding double quotes if we changed all column names to lower case?

VictorVelarde:
Not silly. The problem comes from the use of field names in **user datasets** like "My colum", "Lat point" and alike

packages/custom-storage/__tests__/SQLStorage.spec.ts
```diff
@@ -186,9 +186,9 @@ describe('SQLStorage', () => {
   //     name: 'My Vis',
```
VictorVelarde: let's remove the commented lines, if they are not useful

VictorVelarde:
Ping me when it's ready, so I can rebase at metrics PR

PR #73, opened 2020-03-25
Add source and style update methods

jesusbotella:
This PR adds `.setSource` and `.setStyle` methods to update the current source and style defined in within created layers.

**.setSource(source: string)**
This method accepts a string defining the new CARTO dataset that the layer will render. It creates the new source instance internally and creates a new layer that is passed to Deck.gl to update the map.

**.setStyle(style: {})**
This method accepts an object defining Deck.gl's accessors that the layer will use to define each feature style. It creates the new style instance internally and creates a new layer that is passed to Deck.gl to update the map.


packages/viz/src/lib/Layer.ts
```diff
@@ -75,6 +125,10 @@ export class Layer {
     return { urlTemplates, geometryType };
   }
 
+  private generateLayerId(layerSource: string) {
+    return `MVTLayer-${layerSource}`;
```
VictorVelarde, line 129: How would this work with a long SQL complex query...? Should we create some  hash?

VictorVelarde (approved):
LGTM. Really nice!

PR #75, opened 2020-04-06
[MAP] Create helper functions for basemaps

VictorVelarde:
[ch64278]


packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
+
+/**
+ * Mapping libraries are meant to be available in the browser before using these helpers
+ */
+declare global {
```
jesusbotella: I don't know if this declaration will solve the problem below with Deck.gl's usage from window. In any case, we should move it to a declarations file.

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    google: {
+      maps: {
+        Map(mapContainer?: any, opts?: any): any;
+      }
+    };
+  }
+}
+
+interface StyleUrlCatalog { [key: string]: string; }
```
jesusbotella: I think we should split these definitions and the basemap URL definition from functions to create maps.

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+ *    const deckMap = carto.viz.map('darkmatter', { zoom: 4 });
+ *    const deckMap = carto.viz.map('positron', { zoom: 4, longitude: 3, latitude: 40, pitch: 45, bearing: 30 }, 'map');
+ * ```
+ * @export
+ * @param {string} [basemap='positron']
+ * @param {*} [view]
+ * @param {string} [containerId='map']
+ * @returns
+ */
+export function map(basemap: string = 'positron', view?: any, containerId: string = 'map') {
```
jesusbotella: I think that instantiation like `carto.viz.map` is weird because you might think that you need to add `new` to instantiate a new map like Deck.gl or any other map libraries.

So, as it is not a class, why don't we create a `createMap` function?

When it comes to parameter definition, I think we should admit an object with the properties, because there's no difference in importance to have a main parameter and an options object.

So I think we can make something like this:
```js
const DEFAULT_VIEW = {
  longitude: 0,
  latitude: 0,
  zoom: 1
};

interface MapOptions {
  basemap: string;
  view: {
    longitude: number;
    latitude: number;
    zoom: number;
  },
  container: string;
};

const defaultOptions = {
  container: '#map',
  basemap: 'voyager',
  view: DEFAULT_VIEW
};

export function createMap(options: MapOptions = defaultOptions) {
}
```

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+  const mapStyle = cartoBasemaps[basemap.toLowerCase()];
+
+  const DEFAULT_VIEW = {
+    longitude: 0,
+    latitude: 0,
+    zoom: 1,
+  };
+  const initialViewState = Object.assign({}, DEFAULT_VIEW, view);
+
+  const deckMap = new (window.deck.DeckGL as any)({
```
jesusbotella: We should not need to set Deck.gl as any if we had previously defined types in TS

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+ *    const deckMap = carto.viz.gmap('hybrid', { zoom: 4 });
+ *    const deckMap = carto.viz.gmap('terrain',  { zoom: 4, center: { lng: 3, lat: 40 } }, 'map');
+ * ```
+ * @export
+ * @param {string} [mapTypeId='roadmap']
+ * @param {*} [mapOptions] https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions
+ * @param {string} [containerId='map']
+ * @returns
+ */
+export function gmap(mapTypeId: string = 'roadmap', mapOptions?: any, containerId: string = 'map') {
```
jesusbotella: I think we should do something like what I proposed above in here

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (!window.google.maps.Map) {
+    throw new Error(
```
jesusbotella: We should use CartoError whenever possible

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+      'This helper is meant to be used on the browser, with googlemaps & deck available at window'
```
jesusbotella: Let's respect brand names, and print specific errors for libraries like:
```js
'Google Maps JS library is not found within window context. Please check documentation to know more about this.'
```

packages/viz/src/lib/basemap.ts
```diff
@@ -0,0 +1,112 @@
...
+ * ```
+ * @export
+ * @param {string} [basemap='positron']
+ * @param {*} [view]
+ * @param {string} [containerId='map']
+ * @returns
+ */
+export function map(basemap: string = 'positron', view?: any, containerId: string = 'map') {
+  if (!window.deck.DeckGL) {
+    throw new Error(
```
jesusbotella: Let's respect brand names, and print specific errors for libraries like:
```js
'Deck.gl library is not found within window context. Please check documentation to know more about this.'
```

jesusbotella (commented):
I've pointed out some things that we can improve in our basemap/map definition helpers.

packages/viz/src/lib/basemap.ts
VictorVelarde: It works for these helpers, and yes I'll extract it to another file!

packages/viz/src/lib/basemap.ts
VictorVelarde: 👍 I have simplified this, creating a simple `CartoMapStyle` enum in another file

packages/viz/src/lib/basemap.ts
VictorVelarde: Yes, the idea was to simplify the function, but it's true it might be confusing, regarding to the 'missing' new. I will give it another try, changing name and params options...

packages/viz/src/lib/basemap.ts
VictorVelarde: The type defines an interface but it doesn't define a ctor function... I think we have to keep it by now, or change the window declaration

packages/viz/src/lib/basemap.ts
VictorVelarde: Nice advise! 👍

packages/viz/src/lib/basemap.ts
VictorVelarde: I've modified the messages and created a gfeneric `CartoBaseMapError` for these

packages/viz/src/lib/basemap.ts
VictorVelarde: First step: clearer names --> `createMap`, `createGoogleMap` (both will come inside `carto.viz`)

VictorVelarde:
@jesusbotella I have applied the suggestions and I made a refactor, which lead to clearer helper functions. Have a look at it before we move on

```js
     const deckMap = carto.viz.createMap();
     const deckMap = carto.viz.createMap({ basemap: 'voyager'});
     const deckMap = carto.viz.createMap({ basemap: 'voyager', view: { zoom: 4 } });
     const deckMap = carto.viz.createMap({ basemap: 'positron', view: { zoom: 4, longitude: 3, latitude: 40, pitch: 45, bearing: 30 }, container: 'map' });
```

and

```js
     const deckMap = carto.viz.createGoogleMap();
     const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'satellite'});
     const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'hybrid', zoom: 4 });
     const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'terrain', zoom: 4, center: { lng: 3, lat: 40 } });
```

I updated also the viz examples (which use `createMap` option)

packages/viz/src/lib/basemap/create-map.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+export function createMap(options: MapOptions = DEFAULT_OPTIONS) {
+  if (!window.deck.DeckGL) {
+    throw new CartoBaseMapError(
+      'Deck.gl library not found within window context. Please check documentation to know more about how to configure it.'
+    );
+  }
+
+  const chosenOptions = {
+    basemap: options.basemap || DEFAULT_OPTIONS.basemap || 'positron',
```
jesusbotella, line 51: I think we're over defaulting 😁. I don't think we need to set that `'positron'` in string.

packages/viz/src/lib/basemap/create-map.ts
```diff
@@ -0,0 +1,67 @@
+import { CartoMapStyle } from './CartoMapStyle';
+import { CartoBaseMapError } from '../errors/basemap-error';
+
+interface MapOptions {
```
jesusbotella: I don't know what would happen with interface collisions. I think we have an interface called MapOptions already. I don't know if it'd be better to call this one `DeckGLMapOptions` or something similar.

packages/viz/src/lib/basemap/create-map.ts
```diff
@@ -0,0 +1,67 @@
...
+    );
+  }
+
+  const chosenOptions = {
+    basemap: options.basemap || DEFAULT_OPTIONS.basemap || 'positron',
+    view: { ...DEFAULT_OPTIONS.view, ...options.view },
+    container: options.container || DEFAULT_OPTIONS.container
+  };
+
+  const deckMap = new (window.deck.DeckGL as any)({
```
jesusbotella, line 56: I've tried to fix this using the Deck.gl typings that we imported but it doesn't work because props are not updated. We might need to contribute to Deck.gl typings package.

If we wanted to fix this we would need to set our typings like this:
```js
interface Window {
    deck: {
      DeckGL: typeof Deck;
      GoogleMapsOverlay: GoogleMapsOverlay;
    };
}
```

and it would work. But let's leave it like this for now.

packages/viz/src/lib/basemap/create-google-map.ts
```diff
@@ -0,0 +1,51 @@
...
+ *    const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'satellite'});
+ *    const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'hybrid', zoom: 4 });
+ *    const deckMap = carto.viz.createGoogleMap('map', { mapTypeId: 'terrain', zoom: 4, center: { lng: 3, lat: 40 } });
+ * ```
+ * @export
+ * @param {Element | string} container
+ * @param {google.maps.MapOptions} googleMapsOptions
+ * @returns
+ */
+export function createGoogleMap(
```
jesusbotella: I find the new API definitions much better than the older ones!

One thing that might be nit-picking but would be great is to define a common API parameter definition across both type of maps.

Let me explain, both types of maps have some common attributes like the center, zoom, and so on. And then some specific parameters like `mapTypeId` in this case.

So, why don't we change this API specification to have a common specification for common parameters so that developers just need to change the method if they want to convert their CARTO map to a Google Maps map?

packages/viz/src/lib/basemap/create-google-map.ts
```diff
@@ -0,0 +1,51 @@
...
+  container = 'map',
```
jesusbotella: Do you think that this container parameter would be better suited to be within the options parameter? I don't know if it would be common to define a custom container ID, but options are always needed because developers will want to customize their map center and so on.

jesusbotella (commented):
Just some nit-picking comments

packages/viz/src/lib/basemap/create-map.ts
VictorVelarde: Correct, but that's the only way I found to get no complains on line:
`chosenOptions.basemap.toUpperCase() as keyof typeof CartoMapStyle` and keeping at the same time basemap as an optional parameter in the interface

packages/viz/src/lib/basemap/create-map.ts
VictorVelarde: Yeah, that does no harm, I'll rename it

packages/viz/src/lib/basemap/create-map.ts
VictorVelarde: I aggree.

packages/viz/src/lib/basemap/create-google-map.ts
VictorVelarde: Yeah, I thought about that, but I finally decided to keep each other's interface for simplicity, without any abstraction. I think that conversion is interesting but out of scope right now.

packages/viz/src/lib/basemap/create-google-map.ts
VictorVelarde: In my first shot I had it included in a new, bigger options param, and then I simplified it.
I have some doubts about simplicity from the user point of view. Let me give it a try, and we can finally aggree on it

VictorVelarde:
Finally ready @jesusbotella !

VictorVelarde:
Would you please @jesusbotella approve this one? That way we can unlock the other PR and move on

jesusbotella (approved):

PR #76, opened 2020-04-20
Support to source maps

alasarr:
This PR gives support to source maps on sub packages.

image


jesusbotella (approved):
Looks good!

You could've put `sourceMaps` option to true (https://babeljs.io/docs/en/options#sourcemaps) in our common configuration in `babel.config.js` file and thus you will only have to put it once.

https://github.com/CartoDB/toolkit/blob/web-sdk/bundle/babel.config.js#L1-L4

alasarr:
Yep, it's the best way. Unfortunately, there's a bug https://github.com/babel/babel/issues/5261 that ignores `sourceMaps: true` at babel.config.js

PR #77, opened 2020-04-20
Allow passing of aggregation options to map instantiation methods

juandjara:
This is needed to create point cluster layers in the new visualization library


alasarr (approved):
Looks good to me

jesusbotella (approved):
That looks super good! I've checked all tile aggregation properties just in case and you put them all!

💯

alasarr:
@juandjara a tope!

PR #78, opened 2020-04-23
Watch support

alasarr:
`--verbose` is the trick 😄 

It also forces to a specific version of deck.gl


jesusbotella:
I don't think that `--verbose` is the key 😁. The key is the fact that you changed the watch build from webpack UMD to the ESM that Babel generates, which is the one that we are importing from another modules.

In fact, I would remove the `--verbose` attribute.

VictorVelarde:
Aggree on --verbose. Can you explain why you set fixed versions to deck.gl, what was the problem? (in case we need to create an issue upstream)

alasarr:
I had to add `--verbose` to make it actually refresh, without `--verbose` it didn't work

PR #79, opened 2020-04-24
Move logic of tile instantiation to sources

alasarr:
The idea behind this PR is to move out from Layer.ts the logic related with data sources. 

Why? Because if we want to include another kind of layer sources (as Data Observatory layer or a geojson) we can the layer object to have the same interface for styling, to link with deck, with popup, etc..

So with this idea, we've all the logic related to sources at the sources object instead of the Layer
Pending stuff:
- Check all the examples are working. I've only checked `examples/visualization/index.html` 
- TileLayer concept
- Lint
- Fix Test


alasarr:
BTW, This is my first code in TS, feel free to suggest any typo, improvement...

packages/auth/package.json
```diff
@@ -20,7 +20,7 @@
     "build:esm": "BABEL_ENV=ESM npx babel  --source-maps --config-file ../../bundle/babel.config.js --out-dir dist/esm --extensions '.ts' ./src",
     "build:cjs": "BABEL_ENV=COMMONJS npx babel  --source-maps --config-file ../../bundle/babel.config.js --out-dir dist/cjs --extensions '.ts' ./src",
     "build:umd": "webpack --config webpack.config.js",
-    "build:watch": "webpack --config webpack.config.js --watch",
```
VictorVelarde: Nice! Maybe we can point this PR to https://github.com/CartoDB/toolkit/pull/78 branch, so we deal with watch there and merge this after that one

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+  private _deckLayer: any;
 
-    this._layerSource = new Source(source);
-    this._layerStyles = new Style(styles);
-
-    this._layerOptions = Object.assign({}, defaultMapOptions, mapOptions);
-    this._layerInstantiation = this._mapsClientInstance.instantiateMapFrom(
-      buildInstantiationOptions({ mapOptions: this._layerOptions, mapSource: this._layerSource })
-    );
+  constructor(source: string | CARTOSource, styles = {}) {
```
VictorVelarde: We have lost `LayerOptions` param here, so we broke the use of credentials, as I noticed from failing `Layer.spec.ts`.  Did you ommit that on purpouse for any particular reason? If not we should get them back, and inject it to `Source` ctor.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
   }
 
-  private async _replaceLayer(previousLayerSource: Source) {
-    const newLayer = await this.getDeckGLLayer();
-    const previousLayerId = this.generateLayerId(previousLayerSource.getSourceValue());
+  /**
+   * Replace a layer source
+   * @param previousSource source of the layer to be replaced
+   */
+  private async _replaceLayer(previousSource: Source) {
```
VictorVelarde, line 102: In this particular case, we're passing the whole `previousSource` object, but we are using it just to get its id and find the layer. I find this a bit weird, so I would propose to pass just the id. And having a look at the intention of the code, I would suggest a better naming. 

What about something like...?:

```
  private async _refreshLayer(layerId: string) {
    const newLayer = await this._createDeckGLLayer();

    const deckLayers = this._deckInstance.props.layers.filter(
      (layer: any) => layer.id !== layerId
    );

    this._deckInstance.setProps({
      layers: [
        ...deckLayers,
        newLayer
      ]
    });
  }
```

packages/viz/src/lib/sources/Source.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Credentials } from '@carto/toolkit-core';
+import { MapOptions } from '@carto/toolkit-maps';
+
+export interface sourceOptions {
+  credentials?: Credentials;
+  mapOptions?: MapOptions;
+}
+
+/* Interface definition of source blue print */
+export interface blueprint {
```
VictorVelarde: There are some linter rules about naming, so blueprint so it should be `SourceOptions`, `Blueprint`... but we can review and fix them all once we have https://github.com/CartoDB/toolkit/pull/80 merged in

packages/maps/src/Client.ts
```diff
@@ -17,7 +17,8 @@ export class Maps {
    * @param options
    */
   public async instantiateMapFrom(options: MapOptions) {
-    const { sql, dataset, vector_extent = 2048, vector_simplify_extent = 2048, metadata = {}, aggregation = {} } = options;
+    const { sql, dataset, vector_extent = 2048, vector_simplify_extent = 2048,
```
VictorVelarde: We're going to add a common formatter with prettier, so we all get homogeneous spacing automagically

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
-  public async getDeckGLLayer() {
-    const layerProperties = await this._getDeckGLLayerProperties();
-
-    this._mvtLayerInstance = new MVTLayer(layerProperties);
-    return this._mvtLayerInstance;
-  }
+  /**
+   * Method to create the Deck.gl layer
+   */
+  private async _createDeckGLLayer() {
```
VictorVelarde: We have lost the public method `getDeckGLLayer`. I think we could keep a simple public getter on `this._deckLayer`, so a dev can a reference to the pure deck object on demand

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+  vector_simplify_extent: 2048,
+  metadata: {
+    geometryType: true
+  }
+};
+
+/**  
+ * Implementation of a Source compatible with CARTO's MAPs API
+ * **/
+export class CARTOSource extends Source{
```
VictorVelarde: `CartoSource` instead?

packages/viz/src/lib/Layer.ts
```diff
@@ -112,38 +112,13 @@ export class Layer {
...
-    return { urlTemplates, geometryType };
-  }
-
-  private generateLayerId(layerSource: string) {
-    return `MVTLayer-${layerSource}`;
+  /**
+   * Internal method to auto convert string to CARTO source
+   * @param source source object to be converted
+   */
+  private _buildSource(source: string | CARTOSource) {
```
jesusbotella: This method does not need to be part of the class methods, and can be defined as a standalone function.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
 
 export class Layer {
-  private _mapsClientInstance: Maps;
-  private _credentials: Credentials;
 
-  private _layerSource: Source;
-  private _layerStyles: Style;
-  private _layerOptions: MapOptions;
-  private _layerInstantiation: Promise<MapInstance>;
+  private _source: Source;
```
jesusbotella, line 10: This type is CARTOSource, not just Source. And you can remove the Source import above.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
 
   private _deckInstance: any;
-  private _mvtLayerInstance: MVTLayer<string> | undefined;
-
-  constructor(source: string, styles = {}, options: LayerOptions = {}) {
-    const { mapOptions = {}, credentials = defaultCredentials } = options;
-
-    this._credentials = credentials;
-    this._mapsClientInstance = new Maps(this._credentials);
+  private _deckLayer: any;
```
jesusbotella: This is not an any type, even if you change the name here the type should remain the same.

Why would you change the name? This is an MVT Layer, not a generic Deck layer, as it seems right now.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
       await this._replaceLayer(previousSource);
     }
   }
 
+  /**
+   * Add the current layer to a Deck instance
+   * @param deckInstance instance to add the layer to
+   */
   public async addTo(deckInstance: any) {
+    // get current layers
```
jesusbotella: I don't think these two comments explain more than the code does. I'd remove them.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+    const previousSource = this._source;
 
-    this._layerSource = new Source(source);
-    this._layerInstantiation = this._mapsClientInstance.instantiateMapFrom(
-      buildInstantiationOptions({ mapOptions: this._layerOptions, mapSource: this._layerSource })
-    );
+    this._source = this._buildSource(source);
 
-    if (this._mvtLayerInstance) {
+    if (this._deckInstance) {
```
jesusbotella: You would need to check if the current layer exists, not if deck is initialized, because you are going to replace the layer, not the instance,

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+  /**
+   * Add the current layer to a Deck instance
+   * @param deckInstance instance to add the layer to
+   */
   public async addTo(deckInstance: any) {
+    // get current layers
     const currentDeckLayers = deckInstance.props.layers;
-    const createdDeckLayer = await this.getDeckGLLayer();
+    // create the new layer
+    const layer = await this._createDeckGLLayer();
```
jesusbotella: I find this variable name a little shallow, why didn't you like `createdDeckLayer` or `createdMVTLayer`?

packages/viz/src/lib/Layer.ts
jesusbotella: I had this method split into two parts: one for putting together all layer properties, and another one for creating the actual layer. That way we avoided a one long method as this one is. This is a personal preference but I think it'd look better.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+   * Method to create the Deck.gl layer
+   */
+  private async _createDeckGLLayer() {
 
-  private async _getDeckGLLayerProperties() {
-    const { urlTemplates: data, geometryType } = await this._layerInstantiation.then(this._parseInstantiationResult);
-    const defaultGeometryStyles = defaultStyles[geometryType];
-    const layerId = this.generateLayerId(this._layerSource.getSourceValue());
+    // Get the blueprint of the layer
+    const blueprint = await this._source.blueprint();
```
jesusbotella: I don't really like `blueprint`. Is it a widespread term for what we're doing here?

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
-      this._layerStyles.getProperties()
+    // Build the properties expected by Deck.gl
+    const layerProperties =  Object.assign(
+      { data: blueprint.tileURL, id: this._source.id },
+      defaultStyles[blueprint.geometryType].getProperties(),
+      this._styles.getProperties()
     );
+
+    // Create the Deck.gl instance
+    if (this._source instanceof CARTOSource) {
```
jesusbotella, line 87: I think we should check this at the beginning of the method and fail early without creating the layer, like:

```js
import { CartoError } from '@carto/toolkit-core';

private async _createDeckGLLayer() {
  if (!(this._source instanceof CARTOSource)) {
    throw new CartoError('Unsupported source instance')
  }

  // create layer and so on
}
```

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+  /**
+   * Replace a layer source
+   * @param previousSource source of the layer to be replaced
+   */
+  private async _replaceLayer(previousSource: Source) {
+    const newLayer = await this._createDeckGLLayer();
 
     const deckLayers = this._deckInstance.props.layers.filter(
-      (layer: MVTLayer<string>) => layer.id !== previousLayerId
+      (layer: any) => layer.id !==  previousSource.id
```
jesusbotella: This type is not `any`, it is `MVTLayer<string>` as I previously had.

packages/viz/src/lib/Layer.ts
jesusbotella: Regarding method's name, we're not going to refresh the layer, we're going to replace it

packages/viz/src/lib/Layer.ts
```diff
@@ -1,107 +1,107 @@
...
+    // Get the blueprint of the layer
+    const blueprint = await this._source.blueprint();
 
-    return Object.assign(
-      { data, id: layerId },
-      defaultGeometryStyles.getProperties(),
-      this._layerStyles.getProperties()
+    // Build the properties expected by Deck.gl
+    const layerProperties =  Object.assign(
+      { data: blueprint.tileURL, id: this._source.id },
```
jesusbotella: There's like a convention in Deck.gl to have a prefix with the layer's name. So I think we should leave it like: `MVTLayer-{this._source.id}`

packages/viz/src/lib/Layer.ts
```diff
@@ -112,38 +112,13 @@ export class Layer {
...
-    ...mapOptions,
-    ...mapSource.getSourceOptions()
-  };
-}
-
-interface LayerOptions {
-  credentials?: Credentials;
-  mapOptions?: MapOptions;
-}
+}
```
jesusbotella: Add a blank line here

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+      this._mapInstantiation = this._mapsClient.instantiateMapFrom(this._mapOptions);
+
+    }
+
+
+    /**
+     * Parse a string to guess the type of the source
+     * @param source string to guess the type
+     */
+    private _getType(source: string) {
```
jesusbotella: This method can be a standalone method outside this class as it doesn't use any of the class instance properties

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+    * This methods returns the blueprint of the TileLayer. 
+    * It returns the blueprint of the source:
+    *   - URL of the tiles provided by MAPs API
+    *   - geometryType
+    */
+    public async blueprint(): Promise<blueprint> {
+      return await this._mapInstantiation.then(this._parseInstantiationResult);
+    }
+
+    private _parseInstantiationResult(instantiationData: any): blueprint {
```
jesusbotella: This method can be a standalone method outside this class as it doesn't use any of the instance properties

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+      const metadata = instantiationData.metadata;
+  
+      const urlData = metadata.url.vector;
+      const tileURL = urlData.subdomains.map(
```
jesusbotella: We can change this name because what it contains is an array of tile URL templates, and using this name seems like we're just returning a URL that points to a single tile.

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+    geometryType: true
+  }
+};
+
+/**  
+ * Implementation of a Source compatible with CARTO's MAPs API
+ * **/
+export class CARTOSource extends Source{
+
+    // type of the source.
```
jesusbotella: Given that we're using comments here, let's put them in JSDoc formatting and we'll move a step forward in terms of documentation

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -0,0 +1,90 @@
...
+    private _credentials: Credentials;
+
+    constructor(source: string, options: sourceOptions = {}) {
+
+      const { mapOptions = {}, credentials = defaultCredentials } = options;
+      
+      // set layer id
+      const id = `CARTOSource-${source}`;
+
+      // call to super class
```
jesusbotella: I'd remove all these comments across this PR. The code should be self-explanatory, and these comments are not needed as they're not in this case.

packages/viz/src/lib/Layer.ts
jesusbotella: The credentials belong to the source, not to the layer anymore, as there can be some layers that won't need them. The new CARTOSource supports that, so I think we're ok. We'd need to change the tests though.

jesusbotella (commented):
The approach is good but I think we need to do some changes. I've pointed some of them, and we'll need to pass the code through the new ESLint and Prettier configuration.

packages/auth/package.json
alasarr: Done!

packages/viz/src/lib/Layer.ts
alasarr: Done!

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Why? CARTO is the name of the company

packages/viz/src/lib/Layer.ts
alasarr: No, the type is Source (the base class because ). This variable could storage CARTOSource or DataObservatorySource (or GeoJSONSource in the future)

packages/viz/src/lib/Layer.ts
alasarr: Agree 😄

packages/viz/src/lib/Layer.ts
alasarr: It's not a generic Deck Layer, because it could have MVTLayer for CARTOSource, DataObservatoryLayer for DataObservatorySource and GeoJSONLayer in the future. 

There is an issue with the types. I've tried to add TileLayer, MVTLayer and Layer, but the typing complains because the getPickinfo are differents.

TileLayer getPickingInfo:
```
getPickingInfo({ info, sourceLayer }: {
			info: any;
			sourceLayer: any;
		}): any;
```

Layer: 
```
	getPickingInfo({ info, mode }: {
			info: any;
			mode: any;
		}): any;
```

CompositeLayer:
```
	getPickingInfo({ info }: {
			info: any;
		}): any;
```

packages/viz/src/lib/sources/CARTOSource.ts

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Agree 😄

packages/viz/src/lib/Layer.ts
alasarr: No special reason...

packages/viz/src/lib/Layer.ts
alasarr: Me neither, it a concept I tried but it doesn't work

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: done

alasarr:
Ready for reviwe

packages/viz/src/lib/Layer.ts
alasarr: Done

packages/viz/src/lib/Layer.ts
alasarr: The type of the source already forces it. The type is string | Source

At the constructor, we force to convert from string to CARTOSource, so at this point, the instance of the class is guaranteed to be a source

jesusbotella:
I have not reviewed all new changes because I think that there's something to important to discuss here.

As per the latest changes, we are embedding a new type of layer (DataObservatoryLayer) which inherits from a MVTLayer inside main our Layer class. I think that the mechanism of checking what kind of source is the one that we have and instantiate a layer based on that is not optimal because we can have many if clauses that we don't want to.

So, given that we are refactoring our sources and layers, why don't we allow each source to define the type of layer that it wants to use and have an MVT Layer by default?

I don't like this type of source checks (and I think we should avoid them at all times):
```js
if (this._source instanceof CARTOSource) {
  this._deckLayer = new MVTLayer(layerProperties);
} else if (this._source instanceof DataObservatorySource) {
  this._deckLayer = new DataObservatoryLayer(layerProperties);
} else {
  throw Error('Unsupported source instance');
}
```

For example, if we have a GeoJSON source (and we will eventually have it), we will have to add another if clause to the ones we already have. This can be fixed with a factory, but I think it is smarter if we allow each source to define the type of layer that wants to use.

@VictorVelarde @alasarr

examples/visualization/data-observatory/index.html
```diff
@@ -0,0 +1,80 @@
...
+          },
+          controller: true
+        });
+
+        const DOVariable = 'total_pop_624c2d45';
+        const styles = carto.viz.colorBinsStyle(DOVariable, {
+          bins: [0, 2*(10**6), 5*(10**6), 8*(10**6), 100*(10**6)],
+          colors: 'teal'
+        });
+        const source = new carto.viz.DataObservatorySource(DOVariable);
```
VictorVelarde: I like the API, but just a small comment regarding to naming: we should probably use DOSource, (if we already have DOVariable)

packages/viz/src/lib/Layer.ts
VictorVelarde: I would go for a generic createDeckLayer, as we might change the types of layers in the future

packages/viz/src/lib/Layer.ts
```diff
@@ -1,164 +1,133 @@
...
-    const {
-      urlTemplates: data,
-      geometryType
-    } = await this._layerInstantiation.then(parseInstantiationResult);
+  /**
+   * Method to create the Deck.gl layer
+   */
+  public async _createDeckGLLayer() {
+    // Get properties of the layer
+    const props = await this._source.getLayerProps();
```
VictorVelarde, line 78: I would think about a `getMetadata`, instead of `getLayerProps`. That data could be useful for other things, beyond layer

packages/viz/src/lib/Layer.ts
VictorVelarde: It's just a naming discussion, but a 'replace' keeping the id is the only way to 'refresh' that deckgl exposes, isn't it?

jesusbotella (approved):

VictorVelarde (approved):

PR #80, opened 2020-04-24
Add ESLint and Prettier

jesusbotella:
WIP.


VictorVelarde:
I have added github actions to web-sdk in another PR (already merged, and keeping for the moment CircleCI). Then I cherry picked that into this PR, to see if linter executed on that new CI had a different result... but I'm afraid we still get errors... 😢 . (feel free to delete and push forced if you need)

We probably have a different version of eslint or its 'extra-packages' in the CI, that derives in different rules been applied. Maybe we should used fixed versions for these in the meantime:

```
    "eslint": "^6.8.0",
    "eslint-config-airbnb-base": "^14.1.0",
    "eslint-config-prettier": "^6.11.0",
    "eslint-plugin-import": "^2.20.2",
    "eslint-plugin-jest": "^23.8.2",
    "eslint-plugin-prettier": "^3.1.3",
```

jesusbotella:
It's been hard to debug but I've found the culprit by fine debugging the process.

The thing was that Lerna in my computer was not parsing glob patterns correctly (and thus, it was not checking the whole directory) as they were not inside quotes. Why would that happen? No idea, but that was the solution :(

.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+  "useTabs": false,
```
VictorVelarde, line 2: The other day I read this, and it open my eyes to tabs vs spaces: https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/

PR #82, opened 2020-04-24
Use GitHub actions for CI

VictorVelarde:


VictorVelarde:
Not removing circleci until we're 100% happy with the solution

PR #83, opened 2020-04-28
[#63726] SizeBinsStyle & SizeCategoriesStyle implementation

manmorjim:
CH: https://app.clubhouse.io/cartoteam/story/63726/size-sizebinsstyle-sizecategoriesstyle-manual


packages/viz/__tests__/SizeBinsStyle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+        () => sizeBinsStyle('attributeName', { bins: 'invalid' })
+      ).toThrow();
+    });
+    it('should fail when creating a sizeBinsStyle instance with invalid parameters II', () => {
+      expect(() =>
+        sizeBinsStyle('attributeName', {
+          bins: [0, 10],
+          sizes: [2, 10, 15, 50]
+        })
+      ).toThrow();
```
VictorVelarde: It could be interesting to add an expectation here about  a `CartoError` and add internally some specific messages, regarding in this case to number mismatch

packages/viz/__tests__/SizeBinsStyle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    });
+  });
+
+  describe('getRadius and getLineWidth', () => {
+    it('should returns the radius and line width calculated with all parameters by default', () => {
```
VictorVelarde: typo returns

packages/viz/__tests__/SizeBinsStyle.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+          defaultSizeBinsOptions.nullSize
+        )
+      );
+    });
+    it('should gets the min/max values from all parameters', () => {
+      const sizeBinStyleInstance = sizeBinsStyle('attributeName', {
+        bins: [0, 10, 15, 20],
+        sizes: [5, 15, 30],
+        othersSize: 1,
+        nullSize: 60
```
VictorVelarde: Have you tried using nullSize: 0 ? Does it work? Maybe it could be a first way to filter out unwanted values

packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -0,0 +1,165 @@
...
+  const { bins, sizes, othersSize, nullSize } = extendDefaults(options);
+  validateBinParameters(featureProperty, bins, sizes);
+
+  // Number.MIN_SAFE_INTEGER is here to make closed intervals,
+  // that way last range comparison will never be true
+  const ranges = [...bins, Number.MIN_SAFE_INTEGER];
+
+  /**
+   * @private
+   * Gets the size for the feature provided by paramter
```
VictorVelarde: typo 'paramter'

packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * SizeBinsStyle options for sizeBinsStyle
+ */
+interface SizeBinsStyleOptions {
+  /**
+   * The size classes.
+   *
+   */
+  bins: number[];
```
VictorVelarde: Nice interface! In the future bins will be optional too, after adding the 'method' (quantiles, eq. interval..)

VictorVelarde:
I just made some minor comments. I would like to review it with more detail tomorrow.

One really relevant aspect for this PR is the use of "pixel size" units for Points, so we can be homogeneous between points & lines (and document the units in helpers docs). What are our next steps for this? are we going after a dynamic getRadius based on zoom (for points)?

alasarr:
@manmorjim some conflicts after the big merge :)

packages/viz/__tests__/SizeBinsStyle.spec.ts
manmorjim: good point! I added a test to cover that case also. Yes, by this way we can filter out null values with `nullSize` or features out of the range with `otherSize`

examples/visualization/styles/helpers/size-bins-style.html
```diff
@@ -0,0 +1,79 @@
...
+          container: 'map',
+          mapStyle:
+            'https://basemaps.cartocdn.com/gl/voyager-gl-style/style.json',
+          initialViewState: {
+            longitude: -3.700385991742541,
+            latitude: 40.41532659629741,
+            zoom: 14
+          },
+          controller: true
+        });
```
VictorVelarde: Right now, with map helpers already in web-sdk branch, you can use something like:

```js
const deckMap = carto.viz.createMap({ view: { zoom: 4 } });
```
It will use 'voyager' basemap by default and container: map, but you can add / tweak any parameter

examples/visualization/styles/helpers/size-bins-style.html
```diff
@@ -0,0 +1,79 @@
...
+        await airbnbLayer.addTo(deckMap);
+
+        const distanceStyle = carto.viz.sizeBinsStyle('distance', {
+          bins: [0, 500, 700, 900],
+          sizes: [2, 5, 10]
+        });
+        const retiroWalksLayer = new carto.viz.Layer(
+          'retiro_walks',
+          distanceStyle
+        );
```
VictorVelarde, line 72: This is the https://github.com/CartoDB/cartoframes/blob/v1.0.2/cartoframes/viz/styles/size_bins_style.py equivalent in CF, in case it is useful

examples/visualization/styles/helpers/size-bins-style.html
```diff
@@ -0,0 +1,79 @@
...
+        const airbnbLayer = new carto.viz.Layer(
+          'listings_madrid',
+          availabilityStyle
+        );
+        await airbnbLayer.addTo(deckMap);
+
+        const distanceStyle = carto.viz.sizeBinsStyle('distance', {
+          bins: [0, 500, 700, 900],
+          sizes: [2, 5, 10]
+        });
```
VictorVelarde: This is the https://github.com/CartoDB/cartoframes/blob/v1.0.2/cartoframes/viz/styles/size_bins_style.py equivalent in CF, in case it is useful

packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -0,0 +1,166 @@
...
+ * @private
+ * Add default values to empty options.
+ *
+ * @param options with default values.
+ */
+function extendDefaults(
+  options: SizeBinsStyleOptions = defaultSizeBinsOptions
+) {
+  return {
+    bins: options.bins || defaultSizeBinsOptions.bins,
```
alasarr: It needs to follow CF interface. https://github.com/CartoDB/cartoframes/blob/v1.0.2/cartoframes/viz/styles/size_bins_style.py

alasarr (changes requested):

examples/visualization/styles/helpers/size-bins-style.html
```diff
@@ -52,26 +52,32 @@
...
-        const airbnbLayer = new carto.viz.Layer(
-          'listings_madrid',
-          availabilityStyle,
-          {
-            mapOptions: {
-              bufferSize: {
-                mvt: 30
-              }
+        const airbnbSource = new carto.viz.CARTOSource('listings_madrid', {
+          mapOptions: {
```
VictorVelarde, line 57: It's nice that the source allows passing parameters (especially for advanced users), but if Layer doesn't render well without them, I guess we should hide that logic somehow (for example with defaults per layer geometry type)

Another quetion, why 30? is that pixels?

packages/viz/__tests__/SizeCategoriesStyle.spec.ts
```diff
@@ -134,7 +134,7 @@ describe('SizeCategoriesStyle', () => {
   });
 
   describe('min/max values', () => {
-    it('should get the min/max values from bins', () => {
+    it('should get the min/max values from categories', () => {
       const sizeCategoriesStyleInstance = sizeCategoriesStyle('attributeName', {
         categories: [CATEGORY_1, CATEGORY_2, CATEGORY_3],
         sizes: [5, 15, 30]
```
VictorVelarde: Double check with @alasarr on open/closed intervals, to get the same behaviour between size & color helpers on this

examples/visualization/styles/helpers/size-bins-style.html
manmorjim: Yup, according to the [specification](https://github.com/CartoDB/Windshaft/blob/master/doc/MapConfig-1.8.0.md) that size is in pixels. 

I agree, we should add default values in order to render well the layers. However, we only can add a default value for lines ATM. For points we have to wait for the fix https://github.com/visgl/deck.gl/pull/4557

VictorVelarde: It's gonna be difficult to get different defaults from different geometry types, as these params goes on maps instantation and right now we get the geometry type precisely as the response of that instantiation.

alasarr:
Ready to review @VictorVelarde

examples/visualization/styles/helpers/basic-style.html
```diff
@@ -0,0 +1,64 @@
...
+      const deckMap = carto.viz.createMap({
+        basemap: 'voyager',
+        view: {
+          longitude: -116.846099,
+          latitude: 50.2672297,
+          zoom: 2.5
+        }
+      });
+
+      const basicStyle = carto.viz.basicStyle({
```
VictorVelarde, line 48: Nice helper!

packages/viz/__tests__/Layer.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { Layer as DeckLayer, log } from '@deck.gl/core';
 import { Layer } from '../src/lib/Layer';
+import { defaultStyles } from '../src/lib/style';
```
VictorVelarde, line 3: 👌

packages/viz/src/lib/Layer.ts
```diff
@@ -1,12 +1,12 @@
...
-import { CARTOSource } from './sources/CARTOSource';
-import { DOSource } from './sources/DOSource';
+import { CARTOSource, DOSource } from './sources';
 import { DOLayer } from './deck/DOLayer';
-import { defaultStyles, StyleProperties, Style } from './style';
+import { StyleProperties, Style, defaultStyles } from './style';
+import { LayerStyle } from './style/layer-style';
 
-export class Layer {
+export class Layer implements LayerStyle {
```
VictorVelarde: ✨

packages/viz/src/lib/Layer.ts
```diff
@@ -90,18 +98,18 @@ export class Layer {
     const metadata = this._source.getMetadata();
 
     const styleProps = this._style
-      ? this._style.getProperties(this._source)
+      ? this._style.getProperties(this)
       : undefined;
 
     // Get properties of the layer
     const props = this._source.getProps();
```
VictorVelarde, line 105: should rename that method in the source, to have `const props = this._source.getLayerProps();` ?

packages/viz/src/lib/global-interfaces.ts
```diff
@@ -1,18 +1,11 @@
...
-  Line = 'Line',
-  //   MultiLine = 'MultiLine',
-  Polygon = 'Polygon'
-  //   MultiPolygon = 'MultiPolygon'
-}
+export type GeometryType = 'Point' | 'Line' | 'Polygon';
 
 export interface NumericFieldStats {
-  name: string;
+  name?: string;
```
VictorVelarde: should not the name at least be mandatory?

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -13,6 +13,9 @@ export interface SourceOptions {
 const defaultMapOptions: MapOptions = {
   vectorExtent: 2048,
   vectorSimplifyExtent: 2048,
+  bufferSize: {
+    mvt: 10
```
VictorVelarde, line 22: @manmorjim @alasarr Is this default valid for any type of geometry / style? why 10?

packages/viz/src/lib/sources/Source.ts
```diff
@@ -2,7 +2,11 @@
  *
  * Base Source definition. We should keep here the code shared between different sources
  */
-import { GeometryType, NumericFieldStats, CategoryFieldStats } from '../types';
+import {
+  GeometryType,
+  NumericFieldStats,
+  CategoryFieldStats
+} from '../global-interfaces';
```
VictorVelarde: It's ok to have now that name, but I think we should probably rename that into something more meaningful in the future (maybe exposed them from `./Metadata`?)

packages/viz/src/lib/style/Style.ts
```diff
@@ -1,10 +1,10 @@
 import { GeoJsonLayerProps } from '@deck.gl/layers/geojson-layer/geojson-layer';
 import { CartoStylingError, stylingErrorTypes } from '../errors/styling-error';
-import { Source } from '../sources/Source';
+import { LayerStyle } from './layer-style';
```
VictorVelarde: `StyledLayer` ?

packages/viz/src/lib/style/helpers/style-transform.ts
```diff
@@ -0,0 +1,64 @@
+import { GeometryType } from '../../global-interfaces';
+import { DefaultStyleOptions, defaultStyles } from '..';
+import { hexToRgb } from './utils';
+
+const MAP = {
+  Point: {
```
VictorVelarde, line 6: Nice mapping to ease the transition... 😄

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -112,3 +86,27 @@ export function parseGeometryType(type: string): GeometryType {
...
+    currentIndex: number,
+    valuesArray: number[]
+  ) =>
+    featureValue < definedValue &&
+    (currentIndex === 0 || featureValue >= valuesArray[currentIndex - 1]);
+
+  return buckets.findIndex(rangeComparison);
+}
+
+export function calculateSizeBins(nBuckets: number, sizeRange: number[]) {
```
VictorVelarde, line 105: Why not to put this method into `Classifier`?

VictorVelarde (approved):
LGTM! A really good job here, adding relevant features but also improving the design 👏 
I made some minor suggestions about naming, but feel free to skip them if you wish.

packages/viz/src/lib/style/Style.ts
alasarr: 👌

packages/viz/src/lib/sources/Source.ts
alasarr: Yes, It'm not happy with that solution. In fact, I've changed the name multiple times

alasarr: Done!

packages/viz/src/lib/style/helpers/utils.ts
alasarr: Because the idea of the classifier is just to have a numeric classifier with no extra logic.

packages/viz/src/lib/global-interfaces.ts
alasarr: You catch me :trollface: 

Fixed

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Yes. I added 10 to support because before at the map instantiation we don't know which type of geometry. I'd keep like that and talk if it makes sense to include at maps API a buffer based on the geom type

alasarr (approved):

PR #84, opened 2020-04-28
Adding the bufferSize option to the mapconfig

manmorjim:
This option could be configured at layer level. Example:

```javascript
const mapOptions = {
  bufferSize: {
    mvt: 64
  }
};
const airbnbLayer = new carto.viz.Layer(
  'listings_madrid',
  availabilityStyle,
  {
    mapOptions
  }
);
```

However, according to the [MapConfig specification](https://github.com/CartoDB/Windshaft/blob/47e9e108dd6bb34820af2ea2e29d730bde3e4552/doc/MapConfig-specification.md) this option is especified at `LayerGroup` so maybe it should be defined as map option instead of a layer option.

### DIFF
--- a/packages/maps/src/Client.ts
+++ b/packages/maps/src/Client.ts
@@ -115,7 +115,7 @@ export interface AggregationColumn {
 }
 
 export interface MapOptions {
-  bufferSize: BufferSizeOptions;
+  bufferSize?: BufferSizeOptions;
   sql?: string;
   dataset?: string;
   vectorExtent: number;
--- a/packages/maps/src/Client.ts
+++ b/packages/maps/src/Client.ts
@@ -23,7 +23,8 @@ export class Maps {
       vectorExtent = 2048,
       vectorSimplifyExtent = 2048,
       metadata = {},
-      aggregation = {}
+      aggregation = {},
+      bufferSize
     } = options;
 
     if (!(sql || dataset)) {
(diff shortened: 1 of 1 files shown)


VictorVelarde:
Nice addition!. In carto vl there is a default value of 1 

https://github.com/CartoDB/carto-vl/blob/57a55888a67d8e76bca8a259afaa992722a95c55/src/client/windshaft.js#L333

but that parameter is not exposed externally in the Layer (indeed most of the mapOptions are managed internally, with good defaults). I think this mechanism for specifying it is useful, but it won't probably be used by a lot of (external) users.

 Have you considered setting a default value?

VictorVelarde (approved):
Nice job ✨ !!

manmorjim:
> Nice addition!. In carto vl there is a default value of 1
> Have you considered setting a default value?

I didn't because in the specification is `undefined` but now I'm thinking about the weird behaviors in deckGL we could consider to add a default value that prevent that behavior. I'll check what would be the minimum value which fixes that and add it as default.

manmorjim:
I've tried to check what is the minimum value but I was not able to get it because of the clipping issue described in https://github.com/CartoDB/toolkit/issues/85

Also it would help us get the dynamic radius scale implemented to check the minimum value for a max size 45px (like in builder): https://github.com/CartoDB/toolkit/issues/86

VictorVelarde:
@manmorjim I think we should try somehow to unlink this PR from its related issues. My suggestion is merging this one in, without defaults (we can always decide on this later on). Do you aggree?

VictorVelarde:
We can go ahead and merge this, right?

VictorVelarde:
I'll merge this one, but next time you can merge yourself one it has been approved

PR #87, opened 2020-04-29
Remove package-lock from packages + fix hoisting warnings

VictorVelarde:
I have removed again package-lock files from packages (added them to .gitignore too) as we did in https://github.com/CartoDB/toolkit/pull/54

I also removed  some dependency warnings detected on lerna hoisting, as different versions of webpack were declared in some packages.

Applying the process from scratch, also reorders the packages by name in the package.json files, which is nice.


VictorVelarde:
To be merged after #75

jesusbotella (approved):
Looks good!

PR #90, opened 2020-04-30
Full implementation colorBinsStyle

alasarr:
- [x] ColorBinsStyle
- [x] Classifier equal intervals
- [x] Classifier global intervals
- [x] Classifier stdev

- [x] Adapt to [CARTOframes interface](https://github.com/CartoDB/cartoframes/blob/v1.0.2/cartoframes/viz/styles/color_bins_style.py)
- [x] Support to dynamic classifiers at DO
- [x] SizeBinsStyle blocked by https://github.com/CartoDB/toolkit/pull/83
- [x] Create examples
- [ ] Tests
- [x] Bug with colorRanges. The lowest range was never assigned, all the values at this range were set to otherColors.  

https://app.clubhouse.io/cartoteam/story/61602/classifier-for-helpers-quantiles


packages/viz/src/lib/sources/Source.ts
```diff
@@ -8,6 +8,7 @@ export interface LayerProps {
   geometryType: string;
 }
 
+export type bins = 'quantiles' | 'stdev' | 'equal';
```
manmorjim: I'd name this type as `Classifier` or similar to be more clarifying

```javascript
export type Classifier = 'quantiles' | 'stdev' | 'equal';
```

manmorjim (commented):
LGTM  just one suggestion

packages/viz/src/lib/sources/Source.ts
alasarr: Agree

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -95,10 +95,18 @@ export class CARTOSource extends Source {
     return this._credentials;
   }
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  /* eslint-disable */
+  // @ts-ignore
+  public async bins(
```
VictorVelarde: Not sure if you are going to keep now this approach after what we talk, but just in case: IMO bins should be calculated outside of any Source, as a result of a calculation method in a Quantiles independent function or class. 

Obviously its calculation depends on an input from Metadata (that will come from the Source or, in the future, from client-side Viewport calculations). 

So I vote for these methods in a Metadata class:
- `source.metadata.getSample(1000)``
- `source.metadata.getField('fieldName')

and a mechanism to inject the metadata instance in the styles builder

packages/viz/src/lib/sources/Source.ts
alasarr: Done

PR #91, opened 2020-05-01
Adding Maps API stats to CARTOSource

alasarr:


packages/maps/src/Client.ts
```diff
@@ -114,6 +114,16 @@ export interface AggregationColumn {
   aggregated_column: string;
 }
 
+export interface StatsColumn {
+  topCategories: number;
+  includeNulls: boolean;
+}
+
+export interface Sample {
+  num_rows: number;
```
VictorVelarde, line 123: I commented this in other PR, but I think we must be consistent to avoid errors. If this is a public interface, IMO we should use `numRows` and `includeColumns` and then translate them to what Maps API requires internally

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -33,11 +33,18 @@ export class CARTOSource extends Source {
...
   private _value: string;
-  // Instantiation object
-  private _mapInstantiation: Promise<MapInstance>;
+
   // Internal credentials of the user
   private _credentials: Credentials;
 
+  private _layerProps?: CARTOLayerProps;
+
+  private _fieldStats?: FieldStats;
```
VictorVelarde, line 42: We can assume not that we have just 1 field, but we'll probably change that in the future.

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -33,11 +33,18 @@ export class CARTOSource extends Source {
...
+
+  private _mapConfig: MapOptions;
+
+  private _isMapInstantiate: boolean;
```
VictorVelarde, line 46: `_isMapInstantiated`?

packages/maps/src/Client.ts
alasarr: Make sense, but maybe we can create a separate story for that, we also need to do the same for DO.

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Yes, it could be

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Agree. Let's change

PR #92, opened 2020-05-04
Full implementation ColorCategoriesStyle

alasarr:


packages/maps/src/Client.ts
```diff
@@ -114,6 +114,16 @@ export interface AggregationColumn {
...
 }
 
+export interface StatsColumn {
+  topCategories: number;
+  includeNulls: boolean;
+}
+
+export interface Sample {
+  num_rows: number;
+  include_columns: string[];
```
VictorVelarde, line 124: Should we use camelCase for all interfaces (and map internally to snake_case if required)?

```
numRows: number;
includeColumns: string[];
```

packages/viz/src/lib/Layer.ts
```diff
@@ -74,13 +77,22 @@ export class Layer {
    * Method to create the Deck.gl layer
    */
   public async _createDeckGLLayer() {
+    // The first step is to initialize the source to get the geometryType and the stats
+    const styleField = this._style ? this._style.field : undefined;
```
VictorVelarde: :+1 Using one field is enough now, as our helpers require just one (maybe in the future we would use expressions with several ones, and then we'll add an array)

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -20,7 +22,7 @@ function getSourceType(source: string) {
   return containsSpace ? 'sql' : 'dataset';
 }
 
-interface CARTOLayerProps extends LayerProps {
+interface CARTOLayerProps extends SourceProps {
```
VictorVelarde: what if we rename it to `CARTOSourceProps`?

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -51,36 +60,28 @@ export class CARTOSource extends Source {
...
    * It returns the props of the source:
    *   - URL of the tiles provided by MAPs API
    *   - geometryType
    */
-  public async getLayerProps(): Promise<CARTOLayerProps> {
-    const instantiationData = await this._mapInstantiation;
-
-    const { metadata } = instantiationData;
+  public getLayerProps(): CARTOLayerProps {
+    if (!this.isInitialize) {
```
VictorVelarde: `this.isInitialized`?

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -95,10 +96,108 @@ export class CARTOSource extends Source {
...
+    // Modify mapConfig to add the field stats
+    this._mapConfig.metadata.columnStats = {
+      topCategories: 32768,
+      includeNulls: true
+    };
+    this._mapConfig.metadata.dimensions = true;
+
+    /* eslint-disable @typescript-eslint/camelcase */
+    this._mapConfig.metadata.sample = {
+      num_rows: 1000,
```
VictorVelarde, line 131: In VL we made this  `num_rows` a param, as it was pretty relevant for histograms. But we can leave it as it is for now, and extend later

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -95,10 +96,108 @@ export class CARTOSource extends Source {
...
+      include_columns: [field]
+    };
+    /* eslint-enable @typescript-eslint/camelcase */
+
+    this._mapConfig.aggregation = {
+      columns: {},
+      dimensions: {
+        [field]: { column: field }
+      },
+      placement: 'centroid',
```
VictorVelarde, line 144: Also this 'placement' option had a lot of discussion in the pas. In VL it was fixed, but then we're loosing quite a lot of options supported by our backend. +1 to made it a param in the future

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -95,10 +96,108 @@ export class CARTOSource extends Source {
...
+    // if (this._fieldStats !== undefined) {
+
+    if (field !== undefined) {
+      const columnStats = stats.columns[field];
+
+      switch (columnStats.type) {
+        case 'string':
+          this._fieldStats = {
+            name: field,
+            categories: columnStats.categories
```
VictorVelarde: Are we getting here in categories both 'names' and 'number of elements' per category?

packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -13,7 +13,32 @@ interface Variable {
   name: string;
   slug: string;
   starred: boolean;
-  summary_json: any;
+  summary_json: {
```
VictorVelarde, line 17: We should follow a convention for all of our interfaces on camelCase vs snake_case

packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -13,7 +13,32 @@ interface Variable {
...
+        min_range: number;
+      }
+    ];
+    quantiles: {
+      interquartile_range: number;
+      median: number;
+      q1: number;
+      q3: number;
+    };
+    stats: {
```
VictorVelarde, line 34: It's nice to see that the 'metadata' approach will work also on a DOSource 👏

packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -79,28 +129,59 @@ export class DOSource extends Source {
...
+
+    // Get geography from metadata
+    const geography = this._model.dataset.geography_id;
 
     const geographiesURLTemplate = `${vizURL}/geographies/${geography}/{z}/{x}/{y}.mvt?api_key=${apiKey}`;
     const dataURLTemplate = `${vizURL}/variables/{z}/{x}/{y}.json?variable=${this._variable}&api_key=${apiKey}`;
-    const geometryType = 'MultiPolygon';
 
-    return { geographiesURLTemplate, dataURLTemplate, geometryType };
+    return { type: 'TileLayer', geographiesURLTemplate, dataURLTemplate };
```
VictorVelarde, line 184: Are we loosing `geometryType` here on purpose?

packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -79,28 +129,59 @@ export class DOSource extends Source {
...
   }
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-  // @ts-ignore
-  // eslint-disable-next-line
-  public async getMetadataForField(field: string): Promise<Metadata> {
-    throw new Error('Method not implemented.');
+  public getFieldStats(field: string): NumericFieldStats {
+    if (!this.isInitialize || this._model === undefined) {
+      throw new Error('To run this method you need to first call to init()');
```
VictorVelarde: We can use here specific types of CartoError. We can add them progressively...

packages/viz/src/lib/style/Classifier.ts
```diff
@@ -0,0 +1,137 @@
+import { CartoStylingError, stylingErrorTypes } from '../errors/styling-error';
+import { NumericFieldStats } from '../types';
+
+export type ClassificationMethod = 'quantiles' | 'stdev' | 'equal';
+
+export class Classifier {
```
VictorVelarde, line 6: I thought about having different classes for each classifier, for better testing and isolation. I think we can reuse quite a bit of test specs from VL

packages/viz/src/lib/style/Style.ts
```diff
@@ -1,17 +1,32 @@
...
-    this._styleProperties = { ...styleProperties };
+  constructor(styleProperties: StyleProperties, field?: string) {
+    this._styleProperties = styleProperties;
+    this._field = field;
   }
 
-  public hasProperties() {
-    return Boolean(Object.keys(this._styleProperties).length);
-  }
+  public getProperties(source?: Source) {
```
VictorVelarde, line 18: I think that we could extend the mechanism, maybe using the whole `(layer?: Layer)` as parameter, to allow a more open access to other pieces of info for the style functions. 

Then we could have, internally calls like
`layer.getSource` for stats-related or `layer.getMap().viewport` for zoom-related styles

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -1,34 +1,72 @@
...
-    nullColor = defaultOptions.nullColor,
-    othersColor = defaultOptions.othersColor
-  }: ColorBinsStyleOptions = defaultOptions
+  options?: ColorBinsStyleOptions
+) {
+  const opts = { ...defaultOptions, ...options };
+
+  validateBinParameters(featureProperty, opts.breaks, opts.palette);
+
+  const evalFN = (source: Source) => {
```
VictorVelarde, line 25: 👍 As I said we might consider

```
const layerDependantStyleFunc = (layer: Layer) => {
```

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -1,34 +1,72 @@
...
-  } = getColors(colors, bins.length - 1);
+    othersColor: rgbaOthersColor = hexToRgb(options.othersColor)
+  } = getColors(options.palette, ranges.length);
 
-  const rgbaNullColor = hexToRgb(nullColor);
+  const rgbaNullColor = hexToRgb(options.nullColor);
 
-  const getFillColor = (feature: Record<string, any>) => {
-    const featureValue: number = feature.properties[featureProperty];
+  const getFillColor = (feature: Record<string, any>): RGBAColor => {
```
VictorVelarde, line 76: I guess you're changing this for compatibility, but I always found the way VL worked with bins not very intuitive

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -57,22 +96,30 @@ export function colorBinsStyle(
 function validateBinParameters(
   featureProperty: string,
   values: number[] | string[],
-  colors: string[] | string
+  palette: string[] | string
```
VictorVelarde, line 107: yeah, naming here in the same way looks 👌

packages/viz/src/lib/types.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface NumericFieldStats {
+  name: string;
+  min: number;
+  max: number;
+  avg: number;
+  sum: number;
+  sample?: number[];
+  stdev?: number;
+  range?: number;
```
VictorVelarde, line 18: isn't it derived from mandatory min-max?

VictorVelarde:
I made several 'small' comments, but overall, the main idea LGTM. (explicit source instantiation with metadata +  source-dependant style functions).

Main points for discussion, with things I miss:
- if we're gonna support stats for more than 1 field at the same time (so to allow several combined helpers, eg. color dependant on one field and size for other field). I would vote yes.
- If we extend the mechanism for the new 'style functions', that makes use of the Source, so they can allow a bigger object (Layer).

With probably a bit less of urgency / relevance now:
- to give the source methods a bit more of cohesion (getFieldStats, getGeometryType, etc.) with the use inside of an intermediate Metadata class. The idea is, instead of having smaller 'pieces of state' in the source class, we have a big chunk (this._metadata), updated after instantiation, and the rest of them are functions, extracting pieces from it (eg urls) or calculating (eg from stats)

packages/maps/src/Client.ts
alasarr: Agree, I just did the trick here 😄

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Yes, both

packages/viz/src/lib/sources/DOSource.ts
alasarr: Yes, I didn't realize the existence of CartoError when I wrote this part. Let's do it

packages/viz/src/lib/style/Classifier.ts
alasarr: I think we don't need it right now, so I'll add this extra complexity if this class grows, by the moment I don't see it too large.

packages/viz/src/lib/style/helpers/color-bins-style.ts
alasarr: I changed it because we had a bug. 

For the previous code if you have the following ranges [10,20,30], a value lower than 10 was assigned to a null color instead of the first interval

alasarr:
I agree with the three comments @VictorVelarde. Let's do it, they're not hard right now and they will be hard in the future.

alasarr:
What I don't agree is to adapt to CamelCase the fields of the API. I like the idea to keep them as they are at the API and it requires no extra effort 😄

packages/viz/src/lib/style/helpers/color-bins-style.ts
VictorVelarde: Yes, I think that was on purpose, we thought on defining just closed intervals... in any case, I'm ok if the examples and tests are coherent

packages/viz/src/lib/sources/CARTOSource.ts

packages/viz/src/lib/sources/CARTOSource.ts

packages/viz/src/lib/sources/CARTOSource.ts
alasarr: Done

packages/viz/src/lib/sources/CARTOSource.ts

packages/viz/src/lib/sources/DOSource.ts
alasarr: I just created a specific method getGeometryType outside of layer props, it happens on styles where you only need the geometry type and not the tiles info

packages/viz/src/lib/style/Style.ts
alasarr: I've tried to do it but if we do that we create a circular dependency between Layer and Styles

alasarr:
Completed the implementation of the 2/3 points.

>If we extend the mechanism for the new 'style functions', that makes use of the Source, so they can allow a bigger object (Layer).

I've not done this, because if we do that we create a circular dependency between Layer and Styles. Maybe we could add a map parameter with the instance of the map instead of the layer. Thoughts?

packages/viz/src/lib/style/Style.ts
VictorVelarde: Maybe we can define a new *ILayer* interface, implemented by *Layer*, and make both pieces dependent on it to break the cycle

packages/viz/src/lib/Layer.ts
```diff
@@ -74,13 +77,25 @@ export class Layer {
...
+
+    const metadata = this._source.getMetadata();
+
+    const styleProps = this._style
+      ? this._style.getProperties(this._source)
+      : undefined;
+
     // Get properties of the layer
-    const props = await this._source.getLayerProps();
+    const props = this._source.getProps();
```
VictorVelarde, line 105: IMO we should probably use 'getProps' from 'metadata', but let's move on and plan a refactor on this

packages/viz/src/lib/Layer.ts
```diff
@@ -19,9 +19,12 @@ export class Layer {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private _deckLayer: any | undefined;
 
-  constructor(source: string | Source, styles = {}) {
+  constructor(source: string | Source, style?: Style | StyleProperties) {
     this._source = buildSource(source);
-    this._styles = new Style(styles);
+
+    if (style !== undefined) {
```
VictorVelarde, line 31: If style is undefined we should get 'default styles', so always `buildStyle`right?

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -95,10 +97,115 @@ export class CARTOSource extends Source {
     return this._credentials;
   }
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-  // @ts-ignore
-  // eslint-disable-next-line
-  public async getMetadataForField(field: string): Promise<Metadata> {
-    throw new Error('Method not implemented.');
+  public getMetadata(): SourceMetadata {
```
VictorVelarde, line 100: Nice! ✨

packages/viz/src/lib/sources/DOSource.ts
VictorVelarde: I think geometryType could be part of the Metadata, having just one 'parent' object for all these pieces... (to consider in upcoming PRs, not in this one)

VictorVelarde (approved):
I think we can merge this now. IMO we still have some work to do around source-metadata interaction and some other minors, but this PR is already valuable, and if we don't merge it it's gonna get huge. 

So let's merge and then review with more detail from the integrated branch (refactor, basic tests & examples)

PR #93, opened 2020-05-04
Allow to use multiple layers at same time

manmorjim:
This fixes the error thrown by adding two or more layers to the same
map instance. The error occurred because all layers had the same id.

With this fix, each layer has the id of its source.

### DIFF
--- a/packages/viz/src/lib/sources/CARTOSource.ts
+++ b/packages/viz/src/lib/sources/CARTOSource.ts
@@ -21,11 +21,6 @@ function getSourceType(source: string) {
 }
 
 interface CARTOLayerProps extends LayerProps {
-  /**
-   * id of the layer
-   */
-  id?: string;
-
   // Tile URL template. It should be in the format of https://server/{z}/{x}/{y}..
   data: string | Array<string>;
 }
@@ -85,9 +80,7 @@ export class CARTOSource extends Source {
       'ST_'
     )[1];
 
-    const { id } = this;
-
-    return { id, data: urlTemplate, geometryType };
+    return { data: urlTemplate, geometryType };
   }
 
   public get value(): string {
--- a/packages/viz/src/lib/sources/CARTOSource.ts
+++ b/packages/viz/src/lib/sources/CARTOSource.ts
@@ -21,6 +21,11 @@ function getSourceType(source: string) {
 }
 
 interface CARTOLayerProps extends LayerProps {
+  /**
+   * id of the layer
+   */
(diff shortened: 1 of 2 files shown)


packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -21,6 +21,11 @@ function getSourceType(source: string) {
 }
 
 interface CARTOLayerProps extends LayerProps {
+  /**
+   * id of the layer
```
VictorVelarde: I have some weird 'feelings' about mixing the 'Source' (let's say, the 'model'), with the 'Layer' concept (the 'view'...)

@manmorjim What happens if we use the same Source in 2 different Layers?

manmorjim: I see. Good point :thinking:.
The thing is we have to think about a mechanism that prevents users have to define a id for a layer explicitly because this could be annoying. 

What about using a id made of the _source-id_ and the _style-id_? by this way two different layers with the same source and same style won't be able to be displayed in the map at the same time which IMHO it makes sense since two layers with same source and same style are basically the same layer.

manmorjim: OTOH people from PS have just told me they usually establish an explicit ID in the layer so we have to leave open that possibility also.

VictorVelarde: Manu, I think we should somehow manage this Layer-Id on the Layer side, not here, after the call to get this CARTOLayerProps happens. 

There, externally, we could decide if we generate automatically a layer-id with a 'source' & 'style' hash (by default) or take one id manually passed by the Layer creator

manmorjim: Ok, I agree. I've added a `id` in options for a layer which is generated if no one was provided by the user.

By this way:

```javascript
const airbnbLayer = new carto.viz.Layer(
  'listings_madrid',
  availabilityStyle
);
airbnbLayer.id // 'CARTO-listings_madrid-1588611047881'
```

```javascript
const airbnbLayer = new carto.viz.Layer(
  'listings_madrid',
  availabilityStyle,
  { id: 'test' }
);
airbnbLayer.id // 'test'
```

VictorVelarde: 👌 I think that makes sense. Just double-check with @alasarr also on this

alasarr (approved):

PR #94, opened 2020-05-05
draft vizjson4 examples

alrocar:


examples/vizjson4_dual.json
```diff
@@ -0,0 +1,199 @@
...
+        {
+            "id": "map-0",
+            "type": "map_widget",
+            "layout_id": "map-area-0",
+            "options": {
+                # we need to force a 1:1 relation between layer and widget
+                # widgets are tied to a layer_id, if a layer_id is present in multiple maps, then this could determine to which map the widgets are tied to
+                # other option could be disallow the same layer_id to be present in more than one map
+                "main": true, 
+                "map_provider": "deck.gl",
```
VictorVelarde: it would probably make sense to replace 'map_provider' with something like 'basemap_lib', as we're always gonna use deck.gl, but sometimes we will build over Mapbox GL and other times over GMaps

PR #95, opened 2020-05-05
Fix id

alasarr:


packages/viz/src/lib/Layer.ts
```diff
@@ -105,7 +97,7 @@ export class Layer {
     const props = this._source.getProps();
 
     const layerProperties = Object.assign(
-      this._options,
+      this.id,
```
alasarr, line 100: I know I'm not including an _options object, but I'd add it once we have the first field (not id), right now we only have the id

manmorjim (approved):
LGTM

PR #96, opened 2020-05-06
[#57710] Interactivity & Popups

manmorjim:
API for popups.

CH ticket: https://app.clubhouse.io/cartoteam/story/57710/interactivity-popups

### Add & Remove popup click on a layer

```javascript
// add interaction: popup on click
airbnbLayer.setPopupClick([
  { attr: "host_name", title: "Name" },
  "host_since",
  "country",
  { attr: "host_response_time", title: "Response Time" },
]);

// remove interaction click:
airbnbLayer.setPopupClick(null);
```

### Add & Remove popup hover on a layer

```javascript
// add interaction: popup on hover
airbnbLayer.setPopupHover([{ attr: "host_name", title: "Name" }, "country"]);

// remove interaction hover
airbnbLayer.setPopupHover(null);
```

### Add a simple popup to a map

```javascript
const p = new carto.viz.Popup();
p.setContent("This <b>is</b> a test");
p.setCoordinate([-3.7040553015362803, 40.405442923560564]);
p.addTo(deckMap);
```


packages/viz/src/lib/Layer.ts
```diff
@@ -115,24 +108,35 @@ export class Layer {
     return this._deckLayer;
   }
 
+  private async _getLayerProperties() {
```
alasarr, line 117: 👌

packages/viz/src/lib/Layer.ts
```diff
@@ -115,24 +108,35 @@ export class Layer {
...
+      type: MVTLayer
+    };
+  }
   /**
    * Replace a layer source
-   * @param previousSource source of the layer to be replaced
    */
-  private async _replaceLayer(previousSource: Source) {
-    const newLayer = await this._createDeckGLLayer();
+  private async _replaceLayer() {
```
alasarr, line 133: I like the idea of reusing the layer

packages/viz/src/lib/Layer.ts
```diff
@@ -146,6 +150,139 @@ export class Layer {
...
+      this._options.pickable = !!this._options.onClick;
+    }
+
+    if (this._deckLayer) {
+      await this._replaceLayer();
+    }
+  }
+}
+
+function generatePopupContent(elements: any, featureProperties: any): string {
```
alasarr: Could we move this to popup class?

packages/viz/src/lib/Layer.ts
```diff
@@ -146,6 +150,139 @@ export class Layer {
...
+
+  /**
+   * @public
+   * This method creates popups every time the
+   * user clicks on one or more features of the layer.
+   */
+  public async setPopupClick(elements: PopupElement[] | string[] | null = []) {
+    if (elements && elements.length > 0) {
+      let popup: Popup | null = null;
+      // eslint-disable-next-line padding-line-between-statements, @typescript-eslint/no-unused-vars
```
alasarr, line 162: Add the space and remove the lint disable

packages/viz/src/lib/Layer.ts
```diff
@@ -85,28 +86,20 @@ export class Layer {
...
-      styleProps
-    );
+    const layerProperties = await this._getLayerProperties();
 
     // Create the Deck.gl instance
     if (this._source instanceof CARTOSource) {
-      this._deckLayer = new MVTLayer(layerProperties);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore
+      this._deckLayer = new MapboxLayer(layerProperties);
```
alasarr: If we use this it will break compatibility with Google Maps since Google Maps doesn't offer way to draw a custom Layer. We need to use the native class of Deck.gl

packages/viz/src/lib/Layer.ts
```diff
@@ -1,34 +1,41 @@
...
 import { Source } from './sources/Source';
 import { CARTOSource } from './sources/CARTOSource';
 import { DOSource } from './sources/DOSource';
 import { DOLayer } from './deck/DOLayer';
 import { defaultStyles, StyleProperties, Style } from './style';
+import { Popup } from './popups/Popup';
+import { DeckInstance } from './basemap/create-map';
 
 export class Layer {
+  private _options: LayerOptions = {};
```
jesusbotella: Let's respect constructor arguments order and put this after style property.

packages/viz/src/lib/Layer.ts
```diff
@@ -1,34 +1,41 @@
...
     source: string | Source,
     style: Style | StyleProperties = {},
     options: LayerOptions = {}
   ) {
     this._source = buildSource(source);
     this._style = buildStyle(style);
-    this.id = options.id || `${this._source.id}-${Date.now()}`;
+
+    const defaultId = `${this._source.id}-${Date.now()}`;
+    Object.assign(this._options, options, { id: defaultId });
```
jesusbotella: The linter will warn you about this but we'd rather use the rest operator than Object.assign, so we can make something like:

```js
this._options = {
  id: `${this._source.id}-${Date.now()}`,
  ...options
}
```

packages/viz/src/lib/Layer.ts
```diff
@@ -1,34 +1,41 @@
...
+  }
+
+  public get id() {
```
jesusbotella: Do we need to expose this? It is an internal id

packages/viz/src/lib/Layer.ts
```diff
@@ -67,12 +70,10 @@ export class Layer {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async addTo(deckInstance: any) {
-    const currentDeckLayers = deckInstance.props.layers;
     const createdDeckGLLayer = await this._createDeckGLLayer();
 
-    deckInstance.setProps({
-      layers: [...currentDeckLayers, createdDeckGLLayer]
-    });
+    const map = deckInstance.getMapboxMap();
```
jesusbotella: I think this will break Google Maps functionality, we cannot change this like that. We need to come up with another solution.

packages/viz/src/lib/Layer.ts
```diff
@@ -115,24 +108,35 @@ export class Layer {
     return this._deckLayer;
   }
 
+  private async _getLayerProperties() {
+    const metadata = this._source.getMetadata();
+
+    const styleProps = this._style
+      ? this._style.getProperties(this._source)
+      : undefined;
```
jesusbotella: This will break below usage of `styleProps` if style is undefined. We need to use an empty object so that it doesn't break.

And as far as I see, style will never be a falsy value. It is turned into a class in the constructor and whenever we replace it, so I think it's safe to use `this._style.getProperties(this._source)` below. What do you think?

packages/viz/src/lib/Layer.ts
```diff
@@ -115,24 +108,35 @@ export class Layer {
...
+
+    const props = this._source.getProps();
+
+    return {
```
jesusbotella, line 122: Order matters in this object, so I think we should change it a little bit.

My proposal is to put static properties first (`type` property) so that they can be overriden by dynamic ones below. 
After that I would put `props` from the source, and below that property I would put `this._options` just in case we need to override any properties from the outside.

Styles definition will be the ones below all the previous ones. The ordering in `defaultStyles` and `styleProps` is good.

packages/viz/src/lib/Layer.ts
jesusbotella: Reusing the layer is a bad performance practice from Deck.gl. Here's [the FAQ](https://deck.gl/#/documentation/developer-guide/using-layers?section=faq) that explains the why. So I would leave it like I had it to be better in terms of performance.

packages/viz/src/lib/Layer.ts
```diff
@@ -146,6 +150,110 @@ export class Layer {
...
     return this._source;
   }
+
+  /**
+   * @public
+   * This method creates popups every time the
+   * user clicks on one or more features of the layer.
+   */
+  public async setPopupClick(elements: PopupElement[] | string[] | null = []) {
+    if (elements && elements.length > 0) {
```
jesusbotella, line 166: I think we should move this code to the Popup class class or abstract it in any way.

We are adding concepts that the layer shouldn't be aware of, and IMHO, we should create a Popup instance with methods to create handlers and call them when creating the layer.

packages/viz/src/lib/Layer.ts
```diff
@@ -146,6 +150,110 @@ export class Layer {
...
+      this._options.pickable = !!this._options.onHover;
+    }
+
+    if (this._deckLayer) {
+      await this._replaceLayer();
+    }
+  }
+
+  public async setPopupHover(elements: PopupElement[] | string[] | null = []) {
+    if (elements && elements.length > 0) {
```
jesusbotella, line 192: Same here.

packages/viz/src/lib/Layer.ts
```diff
@@ -146,6 +150,110 @@ export class Layer {
...
+          popup.setContent(popupContent);
+          popup.setCoordinate(lngLat);
+          popup.addTo(this._deckInstance);
+        }
+      };
+
+      this._options.pickable = true;
+    } else {
+      this._options.onClick = undefined;
+      this._options.pickable = !!this._options.onHover;
```
jesusbotella: Be careful with this. If we remove pickability from a layer, highlighting is not going to work if enabled. So we shouldn't change this, and leave it always to the default or the value that the user defined it to.

packages/viz/src/lib/basemap/create-map.ts
```diff
@@ -63,5 +85,7 @@ export function createMap(options: DeckGLMapOptions = DEFAULT_OPTIONS) {
     controller: true
   });
 
+  deckMap.getMapType = () => MapType.MAPBOX_GL;
```
jesusbotella: Adding a custom function to an instance is something that we shouldn't probably do. We need to find a property within the map that defines the kind of map and use it.

packages/viz/src/lib/popups/Popup.ts
```diff
@@ -0,0 +1,138 @@
...
+    this._internalPopup.setContent(this._content);
+    this._internalPopup.addTo(map);
+  }
+
+  /**
+   * Sets the coordinate of the popup.
+   *
+   * @param coordinate with long lat.
+   */
+  public setCoordinate(coordinate: number[]) {
```
jesusbotella: Let's put it in plural :) Coordinates always go in pairs.

jesusbotella (commented):
I pointed out some changes that I would do for performance and usability :)

packages/viz/src/lib/Layer.ts
alasarr: Make senses to me

packages/viz/src/lib/Layer.ts
alasarr: I include it because it could be used externally. For example if you're building a UI to control the layers. For example a layerTree with checkboxes.

packages/viz/src/lib/Layer.ts
alasarr: Agree, I talked about it in the slack channel: https://github.com/CartoDB/toolkit/pull/96/files#r422054110.

I'd go for the native solution. Something like this: https://deck.gl/#/examples/core-layers/geojson-layer-polygons

packages/viz/src/lib/Layer.ts
alasarr: No idea of the perf issues, but agree to keep it if performance get better

packages/viz/src/lib/Layer.ts
manmorjim: yup. You're right! it will never be falsy

packages/viz/src/lib/Layer.ts
manmorjim: Not sure if I get you, do you mean Popup has a method to create a handler and pass it to the Layer constructor? Something like:

```javascript
const clickPopup = new carto.viz.Popup();
const hoverPopup = new carto.viz.Popup({closeButton: false});

const clickHandler = clickPopup.createHandler(['attr_name', 'attr_bar']);
const hoverHandler = hoverPopup.createHandler(['attr_name', 'attr_bar']);

const layer = new carto.viz.Layer('layer_name', { onClick: clickHandler, onHover: hoverHandler });
```

jesusbotella: Yes! That's what I kinda meant with my original comment!

Given that we don't probably want to let the user create those instances outside our Layer class (even though I see no problem in that), we can create those instances within our Layer class when using `setPopupClick` or `setPopupHover` but this way it will be much clearer when reading Layer's code.

packages/viz/src/lib/Layer.ts
```diff
@@ -93,28 +96,20 @@ export class Layer implements StyledLayer {
...
-
-    const metadata = this._source.getMetadata();
-
-    const styleProps = this._style
-      ? this._style.getLayerProps(this)
-      : undefined;
-
-    // Get properties of the layer
-    const props = this._source.getProps();
+    if (!this._source.isInitialized) {
```
alasarr, line 99: When does it happen?

alasarr: I mean, what is the use case to include this condition?

packages/viz/src/lib/Layer.ts
```diff
@@ -93,28 +96,20 @@ export class Layer implements StyledLayer {
...
-      id: this.id,
-      ...props,
-      ...defaultStyles(metadata.geometryType),
-      ...styleProps
-    };
+    const layerProperties = await this._getLayerProperties();
 
     // Create the Deck.gl instance
     if (this._source instanceof CARTOSource) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
alasarr: I think you can remove these comments

packages/viz/src/lib/Layer.ts
```diff
@@ -93,28 +96,20 @@ export class Layer implements StyledLayer {
...
+      // @ts-ignore
       this._deckLayer = new MVTLayer(layerProperties);
     } else if (this._source instanceof DOSource) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
alasarr: And these ones

packages/viz/src/lib/Layer.ts
```diff
@@ -123,20 +118,31 @@ export class Layer implements StyledLayer {
     return this._deckLayer;
   }
 
+  private async _getLayerProperties() {
```
alasarr, line 117: +1000

packages/viz/src/lib/Layer.ts
manmorjim: this is executed every time we change some layer properties. E.g. when we set a new style or a new `onClick` or `onHover` handler the layer is recreated again and that line is executed

packages/viz/src/lib/popups/Popup.ts
```diff
@@ -0,0 +1,282 @@
...
+ */
+export class Popup {
+  private _options: PopupOptions;
+  private _coordinates: number[] | undefined;
+  private _deckInstance: Deck | undefined;
+  private _container: HTMLElement;
+  private _parentElement: HTMLElement | undefined;
+  private _isOpened: boolean;
+
+  constructor(options: any = defaultOptions) {
```
alasarr: Let's create an interface for this:
```typescript
export interface DefaultPopupOptions {
  closeButton: boolean;
  containerClassName: string;
  contentClassName: string;
  closeButtonClassName: string;
}
```

And change the constructor to:
```typescript
constructor(options: Partial<DefaultPopupOptions> = {}) {
```

alasarr: Mmmm I see you already have the interface! Use just the partial and remove the any

packages/viz/src/lib/Layer.ts
alasarr: Ok!

alasarr (changes requested):
Just the partial and ready to go

packages/viz/src/lib/popups/Popup.ts
manmorjim: done! thanks for the tip! :rocket:

alasarr (approved):

PR #97, opened 2020-05-07
Deck-styles

alasarr:


examples/visualization/styles/deck-styles.html
```diff
@@ -0,0 +1,61 @@
...
+        }
+      });
+
+      const styles = {
+        getFillColor: [130,109,186, 229],
+        getLineWidth: 1,
+        getLineColor: [255,255,255,127]  
+      };
+
+      const countriesLayer = new carto.viz.Layer('ne_50m_admin_1_states_1', styles);
```
jesusbotella: How are we handling this styles that are just a plan object? I've just seen that our helpers now return a `Style` class instance.

alasarr: Yes, the styles could a Style class instance or a plain object.

VictorVelarde:
I'm ok with this. Being able to use deckgl styles is gonna be very useful.

---

I have a related-question:

We use custom style builders, our helpers, and we admit also deckgl pure styles, but can't we combine them? I see that as a very useful thing. For example, let's image we have:

```js
const availabilityStyle = carto.viz.colorBinsStyle('availability_365');
```

And we're absolutely happy with the style results, on every aspect but.... point **size**.

I think we need a simple way to pass / override 'extra style' params to helpers, like this:

 ```js
const availabilityStyle = carto.viz.colorBinsStyle('availability_365', { deckStyles: { getRadius: 5, whatEverExtraProp: foo } } );
```
(we could also consider using our own 'styles' set of properties, but I think I would prefer deckgl's)

What do you think?

alasarr:
>I think we need a simple way to pass / override 'extra style' params to helpers, like this:

+1 I agree

VictorVelarde:
I'm gonna merge this as it is, with a bit modified example (https://toolkit-b40dsoarz.now.sh/examples/layer/deck-styles.html), and postpone the discussion about passing 'extra deck-gl' styles as options to helpers. 

cc/ @borja-munoz on that in case he want to consider adding an extra story for that part (**Mixing deck-gl styles with helpers**)

VictorVelarde (approved):
🚀

PR #98, opened 2020-05-07
Top support

alasarr:
I've no included the label, because I've checked with @Jesus89 and it's only used for Legend.

So I think it's a better idea to implement that with the legends because we don't have a mechanism for that yet.


manmorjim (approved):
LGTM!

PR #99, opened 2020-05-07
ColorContinous and SizeContinous

alasarr:


examples/visualization/styles/helpers/basic-style.html
```diff
@@ -46,11 +46,11 @@
       });
 
       const basicStyle = carto.viz.basicStyle({
-        // color: '#f00',
+        color: '#f00',
         size: 10,
```
VictorVelarde, line 50: `size` is not required here (should we create one layer of each type?). 

I would also use lower opacity (eg. 0.5) as the colors you have chosen are quite vivid (and I guess we'll use this as an official example in the future).

examples/visualization/styles/helpers/color-bins-style.html
```diff
@@ -49,12 +49,12 @@
         // breaks: [0, 100, 200, 300, 365],
         // Methods allowed are quantiles, equal and stdev
         method: 'quantiles',
-
-        // opacity: 0.5
-
-        // palette accepts an array of HEX color strings,
-        // or a string pointing to a CARTOColors palette
-        // palette: 'teal'
+
```
VictorVelarde, line 52: I have tried here with bins, and it works if the value is low, but we must cap the max number of bins, because if it is high (eg 20), we run out of colors when getting it from the 'teal' palette and it fails on hexToRgb function

examples/visualization/styles/helpers/color-continuous-style.html
```diff
@@ -0,0 +1,57 @@
...
+  <!-- Map libraries-->
+  <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.8.1/mapbox-gl.js"></script>
+  <script src="https://unpkg.com/deck.gl@8.1.0/dist.min.js"></script>
+
+  <!-- CARTO Deck.gl's Tile Layer -->
+  <script src="/packages/toolkit/dist/umd/index.min.js"></script>
+
+  <script>
+    async function initialize() {
+      carto.setDefaultCredentials({ username: 'cartoframes' });
```
VictorVelarde, line 37: should we always use 'public' user for the data in examples?

examples/visualization/styles/helpers/color-continuous-style.html
```diff
@@ -0,0 +1,57 @@
...
+        basemap: 'voyager',
+        view: {
+          longitude: -116.846099,
+          latitude: 50.2672297,
+          zoom: 2.5
+        }
+      });
+
+      const style = carto.viz.colorContinuousStyle('value');
+      const tempLayer = new carto.viz.Layer('temps', style);
```
VictorVelarde, line 49: I would add at least a palette (perhaps even modify the default for continuous):

```js
const style = carto.viz.colorContinuousStyle('value', { palette: 'oryel' } );
```

examples/visualization/styles/helpers/size-continuous-style.html
```diff
@@ -0,0 +1,57 @@
...
+      carto.setDefaultCredentials({ username: 'cartoframes' });
+
+      const deckMap = carto.viz.createMap({
+        basemap: 'voyager',
+        view: {
+          longitude: -116.846099,
+          latitude: 50.2672297,
+          zoom: 2.5
+        }
+      });
```
VictorVelarde, line 46: Let's adjust the zoom to the data a bit better:

```js
      const deckMap = carto.viz.createMap({
        view: {
          longitude: -124.846099,
          latitude: 38.2672297,
          zoom: 4
        }
      });
```

packages/viz/src/lib/style/helpers/color-continuous-style.ts
```diff
@@ -0,0 +1,75 @@
+import { mix as chromaMix } from 'chroma-js';
```
VictorVelarde: I'm gonna make a change in this file, using chroma.scale vs chroma.mix. I think it will be easier and more efficient.

packages/viz/package.json
```diff
@@ -65,7 +65,9 @@
     "@loaders.gl/mvt": "^2.1.2",
     "@luma.gl/core": "8.1.0",
     "@mapbox/vector-tile": "^1.3.1",
+    "@types/chroma-js": "^2.0.0",
     "cartocolor": "^4.0.2",
+    "chroma-js": "^2.1.0",
```
VictorVelarde, line 70: chroma is really nice. If we eventually use d3 for formatting, we could change to d3-scale, which has a pretty similar API:

```js
var color = d3.scaleLinear()
    .domain([10, 100])
    .range(["brown", "steelblue"]);
color(20); // "#9a3439"
color(50); // "#7b5167"
```

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -9,23 +9,38 @@ export function getUpdateTriggers(accessorFunction: Record<string, unknown>) {
...
+
 export function getColors(
   colorProperty: string | string[],
   colorLength: number
 ) {
   if (typeof colorProperty === 'string') {
-    const { colors: rgbaColors, othersColor } = getColorPalette(
+    const { colors } = getColorPalette(
       colorProperty.toUpperCase(),
       colorLength
```
VictorVelarde, line 38: We should probably manage here any possible exception, trying to get more colors than existing ones in the palette

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -110,3 +125,15 @@ export function calculateSizeBins(nBuckets: number, sizeRange: number[]) {
   };
   return new Classifier(classObj).breaks(nBuckets, 'equal');
 }
+
+const lerp = (x: number, y: number, a: number) => x * (1 - a) + y * a;
+const clamp = (a: number, min = 0, max = 1) => Math.min(max, Math.max(min, a));
+export const invlerp = (x: number, y: number, a: number) =>
+  clamp((a - x) / (y - x));
```
VictorVelarde, line 132: IMO we should move them to some math-utils.ts

VictorVelarde:
Nice work on this!

Have a look at the change I made on continuous helper, to use **chroma.scale** (if you're ok, I think we can move on with this PR).

I also found an error (not related to continuous, but color-bins) if using a high number of color categories.

packages/viz/src/lib/style/helpers/color-continuous-style.ts
alasarr: Cool! Thanks!

examples/visualization/styles/helpers/color-continuous-style.html
alasarr: Yes, any specific dataset for continuous representation? The example of linear interpolation in VL is not so great

VictorVelarde:
If you're in a hurry, go ahead with this

PR #100, opened 2020-05-08
Viewport Features Generator

jesusbotella:
This PR includes our new viewport features generator for MVT layers.

ViewportFeaturesGenerator needs a layer instance and the current Deck.gl instance to determine which features are inside the current viewport by using frustum planes.

API definition is not fully implemented yet, and it may change without prior notice.

Pending tasks
----
- [x] Check corner cases for features
- [x] Feature aggregation by `AVG`, `SUM` and some other options.
- [x] Tests


packages/viz/package.json
```diff
@@ -44,8 +44,8 @@
     "@types/geojson": "^7946.0.7",
     "@types/googlemaps": "^3.39.3",
     "@types/jest": "^24.0.18",
-    "@typescript-eslint/eslint-plugin": "^2.29.0",
-    "@typescript-eslint/parser": "^2.29.0",
+    "@typescript-eslint/eslint-plugin": "^2.31.0",
```
VictorVelarde, line 48: Shouldn't we update all other eslint-plugin entries, in other packages?

packages/viz/tsconfig.json
```diff
@@ -9,7 +9,8 @@
     "experimentalDecorators": true,
     "lib": [
       "dom",
-      "es2015"
+      "es2015",
+      "es2019"
```
VictorVelarde, line 13: Should we use just ESNext? 
Or if we need es2019, shouldn't we remove es2015?

packages/viz/package.json
```diff
@@ -44,8 +44,8 @@
...
+    "@typescript-eslint/parser": "^2.31.0",
```
VictorVelarde, line 49: same for this. I'm gonna do it in a small commit

packages/viz/__tests__/viewport-features/cases/lines/lines.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+const viewportFeaturesResult = {
+  features: [{ street: 'FITZGERALD' }],
+  aggregations: {}
+};
+
+const frustumPlanes: ViewportFrustumPlanes = {
+  near: {
+    distance: 0.0023256811732884296,
```
VictorVelarde, line 28: Nice approach with all those cases in a folder ✨! 

How did you generate these use cases? Did you maybe use a small .html setup, with a known width and height, etc., to observe and validate from real values?  I'm thinking about revisiting this in the future, trying to get some context on how to modify or check a particular spec (and they don't look like obvious to me...) 

Could add some brief notes on them to help a bit?

packages/viz/src/lib/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -0,0 +1,167 @@
...
+  ) {
+    const currentFrustumPlanes = this.getFrustumPlanes();
+    const featureCache = new Set<number>();
+
+    return selectedTiles
+      .map(tile => {
+        const transformationMatrix = getTransformationMatrixFromTile(tile);
+        const features = tile.content || [];
+
+        const featuresWithinViewport = features.filter(feature => {
```
VictorVelarde, line 75: Nice function. It made me think about filtering the features client-side on a condition (like feature.pop_max > 50), and the widgets interaction

packages/viz/__tests__/viewport-features/cases/lines/lines.ts
VictorVelarde: Also, I think that we could probably split the tests in 2 parts: detecting the features in viewport (dealing with these complex geometry computations) and calculating aggregations (much simpler and easier to test, using just a numeric field, and without such a complex mockup). 

What do you think about it?

VictorVelarde (commented):
LGTM. The work is really good. I just made some suggestions regarding to testing strategy.

examples/data/viewport-features.html
```diff
@@ -0,0 +1,66 @@
...
+
+  <div class="as-map-panels">
+    <div class="as-panel as-panel--top as-panel--right as-panel--vertical">
+      <div class="as-panel__element">
+        <section class="as-box">
+          <p class="as-title">Viewport features</p>
+          <p class="as-body">
+            Get the visible features on every viewport change
+          </p>
+          <p id="result" class="as-subheader as-color--success"></p>
```
VictorVelarde, line 36: I found difficult to get the features info with just the console.log, so I decided to bring in an airship panel. If we're ok with such a UI, I can try to update all the examples, and we can then follow the 'template'.

https://toolkit-6dl7w45tl.now.sh/examples/data/viewport-features.html

VictorVelarde (approved):
I think we can probably move on with this one, and discuss a bit later, once we split away the aggregations into dataviews

packages/viz/tsconfig.json
jesusbotella: We cannot use ESNext because we wouldn't have this line converted then:
https://github.com/CartoDB/toolkit/blob/9ad98437694f8580d4a75e6a82cf07094fca94e5/packages/viz/src/lib/interactivity/viewport-features/aggregations.ts#L30

But that said, I think that I would use es2019 to transpile those things and then transpile the rest of the code with Babel.

packages/viz/__tests__/viewport-features/cases/lines/lines.ts
jesusbotella: I got the viewport frustum planes and the rest of the data inspecting and debugging the code with the browser. I'd say that is very unlikely that we need to change them, but nevertheless I would put some instructions just in case.

Given that I am currently splitting both features, viewport features and aggregations, those tests will be split as well. So no problem.

packages/viz/src/lib/interactivity/viewport-features/ViewportFeaturesGenerator.ts
jesusbotella: Yes, that's possible with this!

VictorVelarde:
@jesusbotella are you ok with merging this part now?

jesusbotella:
Yes, going to update the branch and merge it. I think we should address `lib` parameter in tsconfig.json for modules altogether and coordinate it with Babel in another issue.

PR #102, opened 2020-05-11
Alasarr/styles refactor

alasarr:
Styles refactor.

- Partial interface to allow update partial options when using the library in other TS projects. 
- Unify styles definition in a single place. 
- Implement styles per geometry type as CF does
- Copy CF default styles
- Get rid of complex interface inheritance.
- Some fix around several attributes were not working
- Simplify examples


examples/visualization/styles/helpers/basic-style.html
```diff
@@ -46,11 +46,17 @@
       });
 
       const basicStyle = carto.viz.basicStyle({
-        color: '#f00',
-        size: 10,
-        strokeColor: '#00f',
-        strokeWidth: 1,
-        opacity: 1.0
+          // Color: hex, rgb or named color value.
```
simon-contreras-deel: I would use this example to show the default ones, so I would avoid to set any value

packages/viz/src/lib/style/default-styles.ts
```diff
@@ -1,53 +1,121 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// TODO remove any
```
simon-contreras-deel: I would create a ticket for it removing the `TODO`, if not this line could be here for years :)

examples/visualization/styles/helpers/color-bins-style.html
```diff
@@ -45,18 +45,7 @@
...
-        method: 'quantiles',
-        
-        opacity: 0.5,
-        // palette accepts an array of HEX color strings, or a string pointing to a CARTOColors palette
-        palette: 'teal',
-        // Color applied to features which the attribute value is null.
-        nullColor: '#ccc',
-      });
-
+      const availabilityStyle = carto.viz.colorBinsStyle('availability_365');
```
VictorVelarde, line 48: It's nice to see the simplest example.
 
But considering that lots of users will copy & paste, I think we should keep the more verbose one somehow. If not, the user will have to go to the reference to see those extra params.

I tried to do that, in the map helper example, doing this:

```js
      const deckMap = carto.viz.createMap();
      // const deckMap = carto.viz.createMap({ basemap: 'positron'});
      // const deckMap = carto.viz.createMap({ basemap: 'darkmatter', view: { zoom: 4 } });
      // const deckMap = carto.viz.createMap({ basemap: 'positron', view: { zoom: 4, longitude: 3, latitude: 40, pitch: 45, bearing: 30 }, container: 'map' });
```

examples/visualization/styles/helpers/color-categories-style.html
```diff
@@ -45,26 +45,7 @@
...
-        // categories array
-        othersColor: '#FFFFFF10'
-      });
-
-      // roomTypeStyle.getRadius = 100;
-      
-      roomTypeStyle
-
-
+      const roomTypeStyle = carto.viz.colorCategoriesStyle('room_type');
```
VictorVelarde, line 48: same, I would try to display the defaults been applied, so it's easier for the user to 'play' with them

packages/viz/src/lib/style/default-styles.ts
```diff
@@ -1,53 +1,121 @@
...
 
-  getFillColor: hexToRgb('#FFB927'),
-  getLineColor: hexToRgb('#222'),
-  getRadius: 5,
+export function getStyleValue(
+  variable: string,
+  geometryType: GeometryType,
+  options: any
+) {
+  const opts = { ...defaults[geometryType], ...options };
```
simon-contreras-deel: Just a question: Do we just have `Point`, `Line` and `Polygon` in metadata info? I mean, it does not return `multipolygon` for example

examples/visualization/styles/helpers/color-continuous-style.html
```diff
@@ -45,7 +45,10 @@
         }
       });
 
-      const style = carto.viz.colorContinuousStyle('value');
+      const style = carto.viz.colorContinuousStyle('value',{
+        rangeMin: 80,
```
VictorVelarde, line 49: I guess we need to specify them cause helper is still on manual mode, right. Are we already able to get those min / max from the source?

simon-contreras-deel (commented):
Just some fast questions. I was just comparing CF and these ones

examples/visualization/styles/helpers/color-bins-style.html
simon-contreras-deel: I think the opposite :) Since it is a helper example, I will try to show how simple the helper is and the cool result. In case we want to add more, I would do it in an `advance usage examples`

packages/viz/__tests__/Layer.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { Layer as DeckLayer, log } from '@deck.gl/core';
 import { Layer } from '../src/lib/Layer';
-import { defaultStyles } from '../src/lib/style';
+import { getStyles } from '../src/lib/style';
```
VictorVelarde, line 3: Why not to keep `getDefaultStyles`?. I find it more clear

packages/viz/__tests__/SizeBinsStyle.spec.ts
```diff
@@ -1,42 +1,10 @@
...
-      ).not.toThrow();
-    });
-
-    it('should fail when creating a sizeBinsStyle instance with invalid parameters', () => {
-      expect(
-        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
-        // @ts-ignore
-        () => sizeBinsStyle('attributeName', { breaks: 'invalid' })
-      ).toThrow();
-    });
```
VictorVelarde, line 22: Isn't this spec still valid?

packages/viz/__tests__/SizeCategoriesStyle.spec.ts
```diff
@@ -1,14 +1,9 @@
...
-} from '../src/lib/style';
+import { sizeCategoriesStyle } from '../src/lib/style';
 
 describe('SizeCategoriesStyle', () => {
   describe('Style creation', () => {
     it('should create a sizeCategoriesStyle instance properly', () => {
-      expect(() =>
-        sizeCategoriesStyle('attributeName', defaultSizeCategoriesStyleOptions)
-      ).not.toThrow();
+      expect(() => sizeCategoriesStyle('attributeName')).not.toThrow();
```
VictorVelarde: 👍

VictorVelarde: We should try to add more specs, regarding the proper use of bins / method names...

packages/viz/src/lib/style/default-styles.ts
```diff
@@ -1,53 +1,121 @@
...
 import { hexToRgb } from './helpers/utils';
 import { GeometryType } from '../sources/Source';
 
-const pointStyles = {
-  opacity: 1.0,
-  stroked: true,
-  filled: true,
+const defaults: any = {
+  Point: {
+    color: '#EE4D5A',
```
VictorVelarde, line 11: I think we're mixing here default styles with properties that just apply to (some) style helpers. I find it quite awkward. 

IMO those extra 'default' params should be in the style helpers themselves. If we ever need to change something in them, for example the name of a parameter (sizeRange, to size) we would need to change in 2 places: the helper and here

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -5,27 +5,55 @@ import {
...
   CartoStylingError,
   stylingErrorTypes
 } from '../../errors/styling-error';
-import { ColorBinsStyleOptions, defaultColorBinsStyleOptions } from '..';
 import { StyledLayer } from '../layer-style';
-import { toDeckStyles } from './style-transform';
 import { NumericFieldStats, GeometryType } from '../../sources/Source';
+import { getStyleValue, BasicOptionsStyle, getStyles } from '../default-styles';
+
+export interface ColorBinsOptionsStyle extends Partial<BasicOptionsStyle> {
```
VictorVelarde, line 17: Does it mean we will be able to 'override' any style property, for example set the size in the helper itself?

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -5,27 +5,55 @@ import {
...
+
+function defaultOptions(
+  geometryType: GeometryType,
+  options: Partial<ColorBinsOptionsStyle>
+): ColorBinsOptionsStyle {
+  return {
+    bins: 5,
+    method: 'quantiles',
+    breaks: [],
+    palette: getStyleValue('palette', geometryType, options),
```
VictorVelarde, line 40: These are the params that I think should be in this file, and not in defaults. On the other hand, I think they probably don't change much depending on the geometry type

packages/viz/__tests__/SizeBinsStyle.spec.ts
alasarr: Nope, I'm writting right now some tests to cover the helpers

examples/visualization/styles/helpers/color-bins-style.html
alasarr: hahaha, that's funny I did it at https://github.com/CartoDB/toolkit/pull/102/commits/74521fd7538cb4e54c45991b01450c5781530ba7 and I thought it was a little bit overwhelming

examples/visualization/styles/helpers/color-continuous-style.html
alasarr: Yes, if you don't specify it we get the defaults: https://github.com/CartoDB/toolkit/blob/alasarr/styles-refactor/packages/viz/src/lib/style/helpers/color-continuous-style.ts#L11

packages/viz/src/lib/style/helpers/math-utils.ts
```diff
@@ -0,0 +1,11 @@
+const lerp = (x: number, y: number, a: number) => x * (1 - a) + y * a;
+const clamp = (a: number, min = 0, max = 1) => Math.min(max, Math.max(min, a));
+export const invlerp = (x: number, y: number, a: number) =>
+  clamp((a - x) / (y - x));
+export const range = (
```
VictorVelarde, line 16: I understand `clamp` and `invlerp`, but what's `range` in this context ?

packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -1,23 +1,48 @@
...
+
+function defaultOptions(
+  geometryType: GeometryType,
+  options: Partial<SizeBinsOptionsStyle>
+): SizeBinsOptionsStyle {
+  return {
+    bins: 5,
+    method: 'quantiles',
+    breaks: [],
+    sizeRange: geometryType === 'Point' ? [2, 14] : [1, 10],
```
VictorVelarde: 👍 here we have the defaults, what makes more sense to me

packages/viz/src/lib/style/default-styles.ts
alasarr: Yes, the style module only knows about no multi geometries. We convert it from Maps API at https://github.com/CartoDB/toolkit/blob/alasarr/styles-refactor/packages/viz/src/lib/style/helpers/utils.ts#L100

packages/viz/__tests__/SizeCategoriesStyle.spec.ts
alasarr: Yes, agree. Working on that

packages/viz/src/lib/style/helpers/math-utils.ts
alasarr: It's the interpolation between two ranges, it onverts a value from one data range to another

packages/viz/src/lib/style/default-styles.ts
alasarr: Think about the following use case:  the palette property is defined in ColorCategories, ColorBins and ColorContinuous.

The idea behind this is: the const `defaults` at `default-styles` we defined all the styles related with layout that are reused in several components. 

That's true I didn't do it with sizeRange, Just fixed 😄

packages/viz/src/lib/style/helpers/color-bins-style.ts
alasarr: If we do this, the pallete style should be defined in 3 files ColorBins, ColorCategories, ColorContinuous

VictorVelarde:
I like the use of `Partial+, as a way for helper options to be extended with 'basic' style properties. That's a nice addition, and a nice feature to expand the helper use.

On the other hand, I found a bit confusing some parts of the default styles. I think we should extract or at least modify the way in which some of the defaults are organized (specially those belonging to helper params):

For example these could be exported directly 
```js
const NULL_COLOR = '#ccc';
const OTHERS_COLOR = '#777';

const DEFAULT_QUANTITATIVE_PALETTE = 'BluYl';
const DEFAULT_DIVERGING_PALETTE = 'TealRose';
const DEFAULT_QUALITATIVE_PALETTE = 'Bold';
```

or these could go into the size helper itself
```js
const POINT_SIZE_RANGE = [2, 14];
const LINE_SIZE_RANGE = [1, 10];
```

packages/viz/src/lib/style/helpers/math-utils.ts
VictorVelarde: Ok, why don't we give it a more explicit name and a header?

packages/viz/src/lib/style/default-styles.ts
alasarr: In fact, I'm not sure if we want to remove this any. It's pretty straight forward and it's for an internal where we take advantage of the dynamic flexibility of JS. 

If we improve it, I'd like a way to explore how to do this without defining a strong interface

packages/viz/__tests__/Layer.spec.ts
alasarr: Because it gets styles using the values of the user, so it's not just getting the defaults.

Maybe we can move it to another module

examples/visualization/styles/helpers/color-bins-style.html
VictorVelarde: Applying what "Simón says" (clap your hands... 👏 ) could be useful, having 2 examples per helper, one with all defaults, and one / two with more params. Maybe we can ping other people for opinion on this tomorrow

packages/viz/src/lib/style/helpers/color-bins-style.ts
alasarr: Yes! In the previous version, if you were using the library in typescript you had to do:

```javascript
const userOptions = { palette: 'teal'}
const options = { ...defaultColorBinsOptionsStyle, ...userOptions}
const style = carto.viz.colorBinsStyle('temp', options)
```

Now you can just do
```javascript
const style = carto.viz.colorBinsStyle('temp', { palette: 'teal'})
```

examples/visualization/styles/helpers/basic-style.html
VictorVelarde: We already have examples for 'default styles' (lines.html, points.html, polygons.html). This one goes for 'custom' (but basic) styles

packages/viz/src/lib/style/helpers/color-bins-style.ts
VictorVelarde: Why don't we expose the palette (which doesn't depend on geometryType) as a const in that file?

```js
palette: DEFAULT_QUANTITATIVE_PALETTE
```

alasarr:
New version with some of the requested stuff 💪 .

I still need to work on adding more test, but let me know if it looks better

examples/visualization/styles/helpers/color-bins-style.html
alasarr: Make sense to me the idea of advance examples, but let do that in a separate issue.

packages/viz/__tests__/Layer.spec.ts
alasarr: Moved

packages/viz/__tests__/ColorBinsStyle.spec.ts
```diff
@@ -1,10 +1,107 @@
...
+      const style = colorBinsStyle(FIELD_NAME);
+      const response = style.getLayerProps(styledLayer);
+      expect(response).toHaveProperty('getFillColor');
+      expect(response.getFillColor).toBeInstanceOf(Function);
+    });
+  });
+
+  describe('Parameters', () => {
+    it('should launch styling error when bins and breaks missmatch', () => {
+      const style = colorBinsStyle(FIELD_NAME, {
```
VictorVelarde, line 48: Nice tests!

packages/viz/__tests__/ColorBinsStyle.spec.ts
```diff
@@ -1,10 +1,107 @@
...
+      });
+      expect(() => style.getLayerProps(styledLayer)).toThrow(CartoStylingError);
+    });
+
+    it('should launch styling error when breaks and palette missmatch', () => {
+      const style = colorBinsStyle(FIELD_NAME, {
+        breaks: [20, 50],
+        palette: ['#ff0', '#231']
+      });
+      expect(() => style.getLayerProps(styledLayer)).toThrow(CartoStylingError);
```
VictorVelarde, line 60: The more I see the `getLayerProps` call, the more I feel we could rename it to be more clear:

How about?
```ts
  public getStyleProps(styledLayer?: StyledLayer) {
    if (typeof this._styleProperties === 'function') {
      if (styledLayer === undefined) {
        throw new CartoStylingError(
          'No layer instance when calling styles function',
          stylingErrorTypes.SOURCE_INSTANCE_MISSING
        );
      }

      return this._styleProperties(styledLayer);
    }

    return this._styleProperties;
  }
```

packages/viz/src/lib/style/helpers/basic-style.ts
```diff
@@ -1,7 +1,7 @@
 import { StyledLayer } from '../layer-style';
 import { Style, getStyles, BasicOptionsStyle } from '..';
 
-export function basicStyle(options: Partial<BasicOptionsStyle>) {
+export function basicStyle(options: Partial<BasicOptionsStyle> = {}) {
```
VictorVelarde, line 4: So a 100% defaults?

VictorVelarde (approved):
I had another look and 👌 . Tests addition from latests commits LGTM

PR #104, opened 2020-05-13
Fix lines size in size helpers

simon-contreras-deel:
Deck.gl expects different behavior with lines and points:
- points: works with meters, so we need to call `pixel2meters`
- lines: works with pixels and we don't need to call `pixel2meters`

I have changed the 3 size helpers

### DIFF
--- a/examples/visualization/styles/helpers/size-categories-style.html
+++ b/examples/visualization/styles/helpers/size-categories-style.html
@@ -37,20 +37,20 @@
 
     <script>
       async function initialize() {
-        carto.setDefaultCredentials({ username: 'public' });
+        carto.setDefaultCredentials({ username: 'cartoframes' });
 
         const deckMap = carto.viz.createMap({
           basemap: 'voyager',
           view: {
-            longitude: -3.700385991742541,
-            latitude: 40.41532659629741,
-            zoom: 12
+            longitude: 135.70,
+            latitude: -25.48,
+            zoom: 3
           }
         });
 
-        const roomTypeStyle = carto.viz.sizeCategoriesStyle('room_type');
+        const roomTypeStyle = carto.viz.sizeCategoriesStyle('type');
 
-        const airbnbSource = new carto.viz.CARTOSource('listings_madrid', {
+        const airbnbSource = new carto.viz.CARTOSource('roads', {
           mapOptions: {
             bufferSize: {
               mvt: 30
@@ -58,7 +58,7 @@
           }
         });
         const airbnbLayer = new carto.viz.Layer(airbnbSource, roomTypeStyle);
-        airbnbLayer.addTo(deckMap);        
+        airbnbLayer.addTo(deckMap);
       }
 
       initialize();
--- a/packages/viz/src/lib/style/helpers/size-bins-style.ts
+++ b/packages/viz/src/lib/style/helpers/size-bins-style.ts
@@ -114,7 +114,7 @@ function calculateWithBreaks(
     }
 
     const featureValueIndex = findIndexForBinBuckets(ranges, featureValue);
-    return pixel2meters(sizes[featureValueIndex], layerStyle);
+    return sizes[featureValueIndex];
   };
 
   /**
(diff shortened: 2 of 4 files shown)


packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -126,7 +126,7 @@ function calculateWithBreaks(
    * @returns radio size.
    */
   const getRadius = (feature: Record<string, any>) => {
-    return getSizeValue(feature);
+    return pixel2meters(getSizeValue(feature), layerStyle);
```
manmorjim, line 129: We are always returning the size in "pixels" (using the function `pixel2meters` to simulate size in pixels) but I'm not sure whether the user might want the size in meters.

If that could happen then we should check the attribute `radiusUnits` before calling the `pixel2meters`. Something like:

```javascript
if (radiusUnits === 'pixels') {
  return pixel2meters(getSizeValue(feature), layerStyle);
} else {
  return getSizeValue(feature);
}
```

What do you think?

manmorjim (commented):
LGTM! Nice catch! just a question about the radius units

packages/viz/src/lib/style/helpers/size-bins-style.ts
simon-contreras-deel: In fact, I think we are going to need to use pixels instead of meters to follow the same behavior here and CF, but I would want to talk about it later.

About your specific point, `radiusUnits` are fixed by us, so that conditional will always return the same. Are you thinking about changing `radiusUnits` to use `pixels` in lines and `meters` in points too? If yes, it will make sense for me

manmorjim: > About your specific point, `radiusUnits` are fixed by us, so that conditional will always return the same.

Oh, I see, so I should put it as default value instead of fixed.

> Are you thinking about changing `radiusUnits` to use `pixels` in lines and `meters` in points too? If yes, it will make sense for me.

I mean putting the `radiusUnits` as a parameter of the style (with `'pixels'` by default) so by this way the user can specify the points size.

```javascript
// an user who wants size in pixels
const styleInPixels = carto.viz.sizeBinsStyle('availability_365', {
    breaks: [50, 100, 200, 300, 365],
    sizeRange: [2, 10]
});

// an user who wants the size in meters
const styleInMeters = carto.viz.sizeBinsStyle('availability_365', {
    breaks: [50, 100, 200, 300, 365],
    sizeRange: [100, 500],
    radiusUnits: 'meters'
});
```

simon-contreras-deel: ah ok, I see. Nice improvement :) I am going to do that way

alasarr: I don't agree. It adds complexity to the user and we don't have it in CF. Let's keep as it was but with the fixed

examples/visualization/styles/helpers/size-categories-style.html
```diff
@@ -37,28 +37,28 @@
 
     <script>
       async function initialize() {
-        carto.setDefaultCredentials({ username: 'public' });
+        carto.setDefaultCredentials({ username: 'cartoframes' });
```
alasarr, line 40: I suppose you've committed the example by mistake, but I prefer to have the point examples because is more common to use this helper with a point layer rather than lines.

alasarr (changes requested):
Example issue

packages/viz/src/lib/style/helpers/size-bins-style.ts
alasarr: I don't agree. It adds complexity to the user, more work and we don't have it in CF. Let's keep as it's right now.

examples/visualization/styles/helpers/size-categories-style.html
simon-contreras-deel: Not, I did it on purpose. There is no examples with lines and also, this one is the one used in CF for the very same helper

packages/viz/src/lib/style/helpers/size-bins-style.ts
simon-contreras-deel: I am going to keep it as it is right now since we have several opinions about it

examples/visualization/styles/helpers/size-categories-style.html
alasarr: Ok!

alasarr (approved):

PR #106, opened 2020-05-13
Add simple example with onhover

manmorjim:


VictorVelarde (approved):
I'm ok with adding it as it is. Thx!

One thing we should improve is the styling. For that, IMO we should start using what we already have, as much as we can, adding the airship css to the mix, instead of creating new css.

Could we reuse, at least some of the classes, from https://carto.com/developers/airship/reference/#/components/info-window ?

manmorjim:
> I'm ok with adding it as it is. Thx!
> 
> One thing we should improve is the styling. For that, IMO we should start using what we already have, as much as we can, adding the airship css to the mix, instead of creating new css.
> 
> Could we reuse, at least some of the classes, from https://carto.com/developers/airship/reference/#/components/info-window ?

Added to the [CH ticket](https://app.clubhouse.io/cartoteam/story/57710/interactivity-popups#activity-75111)

PR #108, opened 2020-05-13
DataObservatory examples and CDN

alasarr:
- Integrate with the CDN of DataObservatory

- Support to remove a layer


packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -126,6 +129,13 @@ export class DOSource extends Source {
     this._variable = variable;
 
     this._baseURL = `${this._credentials.serverURL}api/v4/data/observatory`;
+
+    this._CDNURL = `https://do-gusc-{domain}.cartocdn.com/USER/api/v4/data/observatory/`;
```
jesusbotella, line 133: Given that this variable is constant we can define it while declaring the variable above. And I guess that we should replace that `USER` in the URL with something else.

packages/viz/src/lib/sources/DOSource.ts
```diff
@@ -126,6 +129,13 @@ export class DOSource extends Source {
     this._variable = variable;
 
     this._baseURL = `${this._credentials.serverURL}api/v4/data/observatory`;
```
jesusbotella, line 131: Shouldn't we always use the CDN URL instead of this one?

packages/viz/src/lib/Layer.ts
```diff
@@ -213,6 +223,23 @@ export class Layer implements StyledLayer {
       await this._replaceLayer();
     }
   }
+
+  public remove() {
+    if (this._deckInstance === undefined) {
+      throw new CartoLayerError(
+        'Cannot remove because it was not attached to map',
```
jesusbotella: I think we should put `This layer cannot be removed because it is not added to a map`.

Just tense changes and using same "language" as we have in our methods 😁

packages/viz/src/lib/errors/layer-error.ts
```diff
@@ -0,0 +1,30 @@
...
+export class CartoLayerError extends CartoError {
+  constructor(message: string, type = layerErrorTypes.DEFAULT) {
+    super({ message, type });
+    this.name = 'CartoLayerError';
+  }
+}
+
+export const layerErrorTypes = {
+  DEFAULT: '[Error]',
+  DECK_MAP_NOT_INSTANTIATED: '[Deck.gl map is not instantiated]',
```
jesusbotella: To follow the same pattern that we have in other error classes, we can put here [Deck.gl Map not found].

The kind of error should not be DECK_MAP_NOT_INSTANTIATED because map can be instantiated but the layer knows nothing about the map, I think we should put DECK_MAP_NOT_FOUND.

packages/viz/src/lib/Layer.ts
```diff
@@ -40,7 +41,10 @@ export class Layer implements StyledLayer {
 
   getMapInstance(): Deck {
     if (this._deckInstance === undefined) {
-      throw Error('Layer not attached to map');
+      throw new CartoLayerError(
+        'Layer not attached to map',
```
jesusbotella: I think we should put `Cannot return map instance because the layer has not been added to a map yet`.

examples/visualization/data-observatory/deckgl-boundaries.html
```diff
@@ -0,0 +1,170 @@
...
+        });
+        await loadGeography(geography);
+      }
+
+      async function loadGeography(g){
+        data_tiles = {};
+        // map.setZoom(g.zoom);
+        variable = g.variable;
+
+        const mvtLayer = new window.deck.MVTLayer({
```
jesusbotella, line 136: Shouldn't we use our DOSource at all times?

jesusbotella (commented):
Just pointed out some comments. There are some JS lines commented out in the examples that we should remove like `// map.setZoom(g)`.

packages/viz/src/lib/sources/DOSource.ts
alasarr: Nope, the CDN is only for viz endpoints not to get the metadata

packages/viz/src/lib/sources/DOSource.ts
alasarr: This issue has been reported to the backend team, but it right now requires to pass the username as a query parameter. Once they'll fix it, we'll replace it.

examples/visualization/data-observatory/deckgl-boundaries.html
alasarr: No because on this example I just wanted to try the boundaries without the merge with data. This is an internal example that we won't make public, it's just to let data team debug the performance of the API.

packages/viz/src/lib/Layer.ts
alasarr: Done

packages/viz/src/lib/Layer.ts
alasarr: Done

packages/viz/src/lib/errors/layer-error.ts
alasarr: Done

jesusbotella (approved):

PR #109, opened 2020-05-13
Update deckgl to 8.1.5 version

manmorjim:
The version `8.1.5` fixes the [clipping issue](https://github.com/visgl/deck.gl/issues/4549)

image


alasarr (dismissed):

alasarr (changes requested):
Building is not working with this update

manmorjim:
The build was failing due to the typescript version `3.9.2` has a bug https://github.com/microsoft/TypeScript/issues/38535 

I've fixed the version `3.8.3` which is prior to `3.9.2`.

VictorVelarde (changes requested):
Nice fix!

Apply the TS update also to each package, please. So all of them have the same pinned version. Once that's done, we can merge this

manmorjim:
TS version updated and pinned for all packages.

I've also updated all examples to use deckgl `8.1.5`

alasarr (approved):

VictorVelarde:
@manmorjim Hey, I was also checking / updating the examples and we need to update ALL deck.gl components to 8.1.5 in package.json files (luma, geo-layers, etc) because otherwise the examples will fail

manmorjim:
> @manmorjim Hey, I was also checking / updating the examples and we need to update ALL deck.gl components to 8.1.5 in package.json files (luma, geo-layers, etc) because otherwise the examples will fail

All deckgl components were updated to `8.1.5` except lumaGL due to deckgl `8.1.5` uses lumaGL version `8.1.2`.

VictorVelarde (approved):

VictorVelarde:
Don't merge this please, I would mix this with my PR on examples

VictorVelarde:
Done in https://github.com/CartoDB/toolkit/pull/110

PR #110, opened 2020-05-14
Update deckgl to 8.1.5 and reorg examples

VictorVelarde:
Takes https://github.com/CartoDB/toolkit/pull/109 work on updates and reorg / simplify examples structure.

With this we will have a plainer structure, just 1 level deep
examples
--- category
      --- example1.html

I also put temporary 'previous' examples, to review, in an examples/old folder

I have also updated references in all examples to latest available:
- MapboxGL 1.10.0
- DeckGL 8.1.5
- GoogleMaps JS 3.39
- (and DeckGL 8.1.5)

If you need to create a new example, we would take `template.html` as a base, in order to ensure same versions. So if you need to bump references, please take care of all the examples and update the template.

As a little helper, there is new `npm run serve` task in package.json, to launch http-server on port 5001


manmorjim (approved):
LGTM!

PR #111, opened 2020-05-14
Fix adding multiple layers without await

VictorVelarde:
Small fix that allows adding several layers without await.

I have also joined the "1 layer per type default styles example"  into /examples/layer/several-layers.html that illustrates it is working.

This example can be useful also for checking default styles are working fine, so that's why I ping you too @oleurud  (it looks like we might need some changes, just compare with: https://carto.com/developers/carto-vl/examples/#example-add-multiple-source-layers)


simon-contreras-deel (approved):

simon-contreras-deel:
Yes, we need to change some sizes and stroke

PR #112, opened 2020-05-14
Publish examples on now project

VictorVelarde:
I have added an integration with`now`, for easier testing. 

Now **view deployment** button on every commit should give us access to online examples.

I have also upgraded the mapping references (mapbox & deck.gl version) on the latest do examples references and moved them to root level, but I didn't test them. Could you please @alasarr check them on the new routes?


alasarr (approved):

alasarr:
That's super cool!!

PR #114, opened 2020-05-18
Fix palette name use in colorCategoriesStyle

VictorVelarde:
Allow using a palette name + an explicit category list in colorCategoriesStyle helper.
It also updates the related example 'set-style' (check: https://toolkit-git-victor-fix-set-style.carto-frontend.now.sh/examples/layer/set-style.html)


VictorVelarde:
@oleurud maybe you can include this in your new block of work for 'params' validation

alasarr (approved):

PR #115, opened 2020-05-19
Style helpers improvements

simon-contreras-deel:
## Plan

By helper and geometry type review:
- Size
- Stroke width
- Stroke color
- Default palette
- size continuous linear extrapolation
- color helpers with lines
- updateTriggers review
- debug examples
- deck.gl examples


VictorVelarde:
## Acceptance checklist

- [x] basic
    - [x] size
    - [x] color
    - [x] strokeWidth
    - [x] strokeColor
    - [x] opacity
    - [x] default styles (no params)
- [x] colorBins (`missing line example`)
    - [x] size
    - [x] strokeWidth
    - [x] strokeColor
    - [x] opacity
    - [x] default palette
- [ ] colorCategories  (`missing line / points examples`)
    - [ ] size
    - [ ] strokeWidth
    - [x] strokeColor
    - [x] opacity
    - [x] default palette
- [ ] colorContinuous (missing lines example)
    - [x] size
    - [x] strokeWidth
    - [x] strokeColor
    - [x] default palette
- [ ] sizeBins
    - [x] color
    - [x] strokeWidth
    - [x] strokeColor
    - **Size 'jumps' on zoom changes, and sometimes it doesn't trigger a redraw and size looks wrong**.
- [ ] sizeCategories
    - [x] color
    - [x] strokeWidth
    - [x] strokeColor
    - [x] default range
    - **Size 'jumps' on zoom changes, and sometimes it doesn't trigger a redraw and size looks wrong**.
- [ ] sizeContinuous
    - [x] color
    - [x] strokeWidth
    - [x] strokeColor
    - **Size jumps / wrong size**

VictorVelarde:
Nice work on this. Debug examples you have prepared are being really useful for validation.

I made a first review of all the debug examples, and everything looks fine except the **size helpers**. I still see 'jumps' and wrong pixel sizes (sizeBins, sizeCategories and sizeContinuous). Logic for adjusting based on viewport is not yet working fine

The rest of tests I made look fine: default palettes, override style options...

simon-contreras-deel:
>  I still see 'jumps' 

In the basic helper, as all the points have the same size, I changed it to use pointRadiusMinPixels  and pointRadiusMaxPixels with the same value: the expected point size. That way, size does not change over zoom (it can not change)

I tried to follow a smilar idea with this cases, but it is not possible. All the stuff we can to to force size, it does not depends on features, but are general to the layer. The only way I can see is to change deck.gl to support pixels width in points.

> wrong pixel sizes

I am checking it.

VictorVelarde:
As we said, let's focus on wrong side (a-must), and forget a bit about 'jumps' (nice-to-have).

packages/viz/src/lib/style/helpers/size-continuous-style.ts
```diff
@@ -143,3 +143,13 @@ function calculate(
     updateTriggers: { getRadius, getLineWidth }
```
VictorVelarde, line 156: I think this updateTriggers won't be effective, as we don't specify a viewport-related variable, right?
cc/ @jesusbotella

packages/viz/src/lib/style/helpers/size-categories-style.ts
```diff
@@ -67,6 +67,11 @@ export function sizeCategoriesStyle(
       const stats = meta.stats.find(
         c => c.name === featureProperty
       ) as CategoryFieldStats;
+
+      if (!stats.categories || !stats.categories.length) {
+        throw new CartoStylingError('The featureProperty has not cagegories');
```
jesusbotella: Typo in categories :)

packages/viz/src/lib/style/helpers/size-continuous-style.ts
jesusbotella: Yes, that update trigger will reevaluate styles when getRadius function reference changes (i.e. when updating styles definition).

If we want it to change, we will need to link it to a variable that changes when the points' size should change. We need to be careful linking triggers to viewport changes because reevaluating styles when panning the viewport without data change may not be ideal.

VictorVelarde: In this particular case, I think the variable should be. **zoomLevel** indeed, not any other viewport change (panning should not trigger)

simon-contreras-deel: `getRadius` is using `pixel2meters` function which is using layer to get the viewport and then, `metersPerPixel` which is finally used to calculate the size on the current zoom. Also, when the user changes the zoom, the function is called.

Do you think there is something wrong?

simon-contreras-deel:
After playing with the @jesusbotella deck example, I see the `pointRadiusScale` does not affect to `pointRadiusMinPixels` and `pointRadiusMinPixels`. So:
- we should not use together or 
- we should apply the `pointRadiusScale` to `pointRadiusMinPixels` and `pointRadiusMinPixels`

I a going to follow the second one if possible.

examples/_debug/deck.gl/example.html
```diff
@@ -0,0 +1,97 @@
...
+      }
+    </style>
+  </head>
+
+  <body>
+    <section id="map"></section>
+    <div id="tooltip" style="position: absolute; z-index: 1; pointer-events: none; background-color: #FFF; padding: 8px;"></div>
+
+    <!-- Map libraries-->
+    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.0.0/mapbox-gl.js"></script>
```
VictorVelarde, line 29: we're using v1.10.0 from mapbox

packages/viz/src/lib/style/helpers/utils.ts
```diff
@@ -93,7 +93,7 @@ export function hexToRgb(hex: string): RGBAColor {
     ];
   }
 
-  throw new Error();
+  throw new Error(`Error parsing hexadecimal color: ${hex}`);
```
VictorVelarde, line 96: We have custom cartoError in case you want to try with it

VictorVelarde (approved):
Nice advances @oleurud 🚀 

Let's move on with it and merge it. 
Then we can create another story in clubhouse for further fixes, in a (smaller) PR

alasarr:
Needs to fix conflicts after merge interactivity

packages/viz/src/lib/style/helpers/utils.ts
simon-contreras-deel: I will take care of it in the validations PRs

PR #116, opened 2020-05-21
Style helper validations

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/76421/validations-of-style-helper-params


packages/viz/__tests__/styles/ColorBinsStyle.spec.ts
```diff
@@ -49,15 +49,73 @@ describe('ColorBinsStyle', () => {
...
+      try {
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
+    });
+
+    it('should fails with invalid bins', () => {
+      const style = colorBinsStyle(FIELD_NAME, {
+        bins: 0
```
VictorVelarde, line 75: This is controlled nicely. IMO we should now do the same for the **upper limit**.

I have just tried with `style-helpers/color-bins-style.html` and bins: 10 and it fails, because we're running out of colors.

Having a look at cartocolors ramps length (quantitative ramps, valid for numerical fields), a proper limit can be set on 7 (included). In Builder, valid number of buckets is indeed 2 to 7.

So let's try to control that too

packages/viz/__tests__/styles/ColorCategoriesStyle.spec.ts
```diff
@@ -62,6 +67,54 @@ describe('ColorCategoriesStyle', () => {
         CartoStylingError
       );
     });
+
+    it('should fails with invalid top', () => {
+      const style = colorCategoriesStyle(FIELD_NAME, {
+        top: 0
```
VictorVelarde, line 73: Using same reasoning as above, we should probably set the maximum top value allowed and be consistent on it. 

/style-helpers/color-categories-style.html doesn't fail with a high number of top, but it has only a low number of variations for the field being used. 

/examples/_old/styles/barcelona.html can be better for testing that limit. I made a quick test with top: 30 on it and it doesn't fail, but I think it does not properly assign colors, so IMO we should cap it too.

The qualitative palettes from cartocolors being used have a maximum length of 12 colors, so I guess that should be the limit too (valida range: 1 to 12)

packages/viz/__tests__/styles/ColorContinuousStyle.spec.ts
```diff
@@ -43,6 +44,45 @@ describe('ColorContinuousStyle', () => {
     });
   });
 
+  describe('Paramters', () => {
```
VictorVelarde, line 47: typo 'Parameters'

packages/viz/__tests__/styles/ColorContinuousStyle.spec.ts
```diff
@@ -43,6 +44,45 @@ describe('ColorContinuousStyle', () => {
...
+    it('should fails with invalid palette', () => {
```
VictorVelarde: typo --> 'should fail'

packages/viz/__tests__/styles/ColorContinuousStyle.spec.ts
```diff
@@ -43,6 +44,45 @@ describe('ColorContinuousStyle', () => {
...
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
+    });
+  });
+
   describe('Data validation', () => {
     const opts = {
       palette: ['#f00', '#00f'],
```
VictorVelarde: Not sure about testing so explicitly with chromaScale in this part of the tests, but nevermind. In any case, I'm gonna add a small fix to allow the use of all the colors in the palette, not just first-last, which I think improves quite a bit the result

VictorVelarde: done --> https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/color-continuous-style.html

VictorVelarde:
🔝  commit removes custom buffer (previously mvt: 30) for both size-categories and size-bins helpers examples. I was not able to see any glitch removing them, and I think examples should be simple. We might add a custom map instantiation example in the future though, passing extra parameters... but I think that's not required yet

I think these examples were created by @manmorjim . Manu can you spot any style issue on them?
- [x]  https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/size-bins-style.html
- [x] https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/size-categories-style.html

packages/viz/__tests__/styles/ColorBinsStyle.spec.ts
simon-contreras-deel: oki

packages/viz/src/lib/style/helpers/size-bins-style.ts
```diff
@@ -177,7 +178,17 @@ function calculateWithBreaks(
...
+  geometryType: GeometryType
+) {
+  if (geometryType === 'Polygon') {
+    throw new CartoStylingError(
+      "Polygon layer doesn't support sizeCategoriesStyle",
+      stylingErrorTypes.GEOMETRY_TYPE_UNSUPPORTED
+    );
+  }
+
   if (options.bins < 1) {
```
VictorVelarde: I think we should also limit the upper bins option. We could potentially have an infinite number of bins, not as in colors (limited by palettes), but I think we should be homogeneous on this. 

So I propose limiting the upper to 7 too. In any case, having more differences is probably non-sense and really difficult for the eye to perceive... (and if anyone wants more 'variation' he can always use size-continuous...)

packages/viz/__tests__/styles/SizeBinsStyle.spec.ts
```diff
@@ -62,7 +62,72 @@ describe('SizeBinsStyle', () => {
...
-      expect(() => style.getLayerProps(styledLayer)).toThrow(CartoStylingError);
+
+      try {
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
+    });
+
+    it('should fails with invalid bins', () => {
```
VictorVelarde: let's limit the upper too (I propose 7 bins). See my comments 🔝

packages/viz/__tests__/styles/ColorContinuousStyle.spec.ts
VictorVelarde: I'm gonna do this myself

manmorjim:
> :top commit removes custom buffer (previously mvt: 30) for both size-categories and size-bins helpers examples. I was not able to see any glitch removing them, and I think examples should be simple. We might add a custom map instantiation example in the future though, passing extra parameters... but I think that's not required yet
> 
> I think these examples were created by @manmorjim . Manu can you spot any style issue on them?
> 
> * [x]   https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/size-bins-style.html
> * [x]  https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/size-categories-style.html

They both look great. I think the MVT buffer is not needed since the issue about clipping was fixed. 

BTW I experimented a very uncommon issue about weird lines in the [size-categories-style.html](https://toolkit-git-style-helper-validations.carto-frontend.now.sh/examples/style-helpers/size-categories-style.html) example. In some zoom levels some _'nonexistent'_ lines appear:

weird_lines

weird_lines_2

I think this is a deckgl issue rather than web-sdk.

VictorVelarde:
Thx @manmorjim. The artifacts might be related to https://github.com/visgl/deck.gl/issues/4604, but we should consider it outside of this PR

simon-contreras-deel:
Added upper limits for bins and top categories in related helpers

packages/viz/__tests__/styles/ColorCategoriesStyle.spec.ts
```diff
@@ -62,6 +67,66 @@ describe('ColorCategoriesStyle', () => {
...
+      });
+
+      try {
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
+    });
+
+    it('should fails with top greather than 12', () => {
```
VictorVelarde: typo 'should fails' --> 'should fail' (search and replace, just in case, cause we had a lot of them)

packages/viz/__tests__/styles/SizeBinsStyle.spec.ts
```diff
@@ -1,9 +1,9 @@
...
-import * as mapsResponse from './data-mocks/maps.number.json';
-import { CARTOSource } from '../src';
-import { ClassificationMethod } from '../src/lib/utils/Classifier';
-import { CartoStylingError } from '../src/lib/errors/styling-error';
+import { sizeBinsStyle, defaultStyles } from '../../src/lib/style';
+import * as mapsResponse from '../data-mocks/maps.number.json';
+import { CARTOSource } from '../../src';
+import { ClassificationMethod } from '../../src/lib/utils/Classifier';
+import { CartoStylingError } from '../../src/lib/errors/styling-error';
 // import { CartoStylingError } from '../src/lib/errors/styling-error';
```
VictorVelarde: remove comment

packages/viz/__tests__/styles/SizeBinsStyle.spec.ts
```diff
@@ -62,7 +61,84 @@ describe('SizeBinsStyle', () => {
...
+      });
+
+      try {
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
+    });
+
+    it('should fail with bins greather than 8', () => {
```
VictorVelarde: 'greater than 7' right?

packages/viz/__tests__/styles/SizeBinsStyle.spec.ts
```diff
@@ -62,7 +61,84 @@ describe('SizeBinsStyle', () => {
...
+        nullSize: -1
+      });
+
+      try {
+        style.getLayerProps(styledLayer);
+      } catch (error) {
+        expect(error).toBeInstanceOf(CartoStylingError);
+      }
     });
   });
```
VictorVelarde, line 144: There is a block of commented code from LIN 145 to 158 that can be deleted...

packages/viz/src/lib/style/helpers/color-categories-style.ts
```diff
@@ -62,6 +63,10 @@ export function colorCategoriesStyle(
         c => c.name === featureProperty
       ) as CategoryFieldStats;
       categories = stats.categories.map((c: Category) => c.category);
+
+      if (!categories.length) {
+        throw new CartoStylingError('Current dataset has not categories');
```
VictorVelarde: "Current dataset has not categories metadata for " + featureProperty   ?

VictorVelarde (approved):
LGTM. Nice job on validators. I made several comments, but you've already faced most of them. Feel fre to merge in when you're done.

PR #118, opened 2020-05-22
DataViews & CategoryDataView

jesusbotella:
This branch implements DataViews, a unified way to get aggregations from data in a source or a layer in the map.

DataViews only accept a Layer for now, so if a Source is provided it will throw.

## DataViews
A dataview is instanciated with a `Layer` or `CARTOSource`, and the column you want to get data from.

```js
const layer = new carto.viz.Layer('my_table');
const dataView = new DataView(layer, 'my_column');
```

Once you have an instance, you can call available methods using the instance.

### .aggregate(aggregations)
This method returns an object with the aggregations requested as object properties. There are 6 possible aggregations: `avg`, `min`, `max`, `count`, `sum` and `percentile_XX`.

Here's an example:
```js
const layer = new carto.viz.Layer('my_table');
const dataView = new DataView(layer, 'my_column');

dataView.aggregate('avg', 'min', 'count', 'sum', 'percentile_90')
```

It will return:
```js
{
  avg: 30,
  count: 5,
  max: 50,
  min: 10,
  sum: 150,
  percentile_90: 50
}
```

### .groupBy(column [, {operation: 'avg' } ])
This methods groups data by the values in the specified colum and aggregates it by the specified operation (`avg` operation by default).

Here's an example:
```js
const layer = new carto.viz.Layer('my_table');
const dataView = new DataView(layer, 'my_column');

dataView.groupBy('column', { operation: 'sum' })
```

It will return:
```js
[
  {name: 'Country 1', value: 2},
  {name: 'Country 2', value: 3}
]
```


packages/data/src/lib/aggregations/aggregations.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+
+  [AggregationTypes.SUM](values: number[]) {
+    return values.reduce((sum, value) => sum + value, 0);
+  },
+
+  [AggregationTypes.PERCENTILE](values: number[], aggregationData: string[]) {
+    const percentile = parseInt(aggregationData[0], 10);
+
+    if (!Number.isInteger(percentile)) {
```
VictorVelarde: We can also add the 0...100 range validation

jesusbotella: Sure!

examples/dataviews/widget.html
```diff
@@ -0,0 +1,65 @@
...
+
+<body>
+  <section id="map"></section>
+
+  <as-category-widget class="as-p--16" heading="Cities" description="Number of countries by continent"
+    style="position: absolute; right: 10px; top: 10px; height: 368px;">
+  </as-category-widget>
+
+  <!-- Map libraries-->
+  <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.8.1/mapbox-gl.js"></script>
```
VictorVelarde: 1.10.0

examples/dataviews/widget.html
```diff
@@ -0,0 +1,65 @@
...
+  <script src="https://unpkg.com/deck.gl@8/dist.min.js"></script>
```
VictorVelarde: IMO we should pin it to https://unpkg.com/deck.gl@8.1.5/dist.min.js
It's more tedious, but that way we share exactly the same version everywhere, and we control the 'upgrade' step (eg. next one soon 8.1.7)

examples/dataviews/widget.html
```diff
@@ -0,0 +1,65 @@
...
+  <script src="https://libs.cartocdn.com/airship-components/v2.1.1/airship.js"></script>
```
VictorVelarde: Same, v2.3 without specifying patch

VictorVelarde: I'm gonna update the 'template.html'.file in this PR, to use it as a reference for examples

packages/data/src/lib/data-view/data-view.ts
```diff
@@ -0,0 +1,132 @@
+import { CartoError, WithEvents } from '@carto/toolkit-core';
+import { Layer, CARTOSource } from '@carto/toolkit-viz';
+import { groupValuesByAnotherColumn } from '../aggregations/grouping';
+import {
+  applyAggregations,
+  AggregationTypes
+} from '../aggregations/aggregations';
+
+export class DataView extends WithEvents {
+  private source: CARTOSource | Layer;
```
VictorVelarde: Thinking about 'source' name... Should we change it to something that accommodates better the 'Layer' option? Maybe **origin** ? So to avoid using the word 'source' that might be associated with just a CARTOSource

packages/data/src/lib/data-view/data-view.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    if (this.source instanceof Layer) {
+      this.source.on('viewportLoad', () => {
+        this.onDataUpdate();
+      });
+    }
+  }
+
+  private onDataUpdate() {
+    this.emitter.emit('dataUpdate');
```
VictorVelarde: Should we go after 'dataUpdated' instead?. We had 'dataChanged' in carto.js

packages/data/src/lib/data-view/data-view.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  async aggregate(...operations: AggregationTypes[]) {
```
VictorVelarde: I know we talked about this design, but what's the main advantage of being able to provide different operations on every call...? I mean. should we simplify it and just pass it as options to the constructor? That's the carto.js model, and of course we can change it but... Should the same dataview change its aggregation ops during its life cycle? In what cases? I can't think of any example...

I guess a similar argument could be held on groupBy.

Sorry if I digress a bit but to me the dataview already implies the concept of 'aggregation', out of the box, and I just find it a bit 'empty' when we create one without specifying the 'aggr. method'

packages/data/src/lib/data-view/data-view.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+      layer.emit('viewportLoad');
+
+      expect(dataUpdateSpy).toHaveBeenCalled();
+    });
+  });
+
+  describe('aggregate', () => {
+    it('should return an empty object if operations are not provided', () => {
+      const dataView = new DataView(new Layer('fake_source'), 'fake_column');
+      expect(dataView.aggregate()).toMatchObject({});
```
VictorVelarde: I'm not sure this use case makes sense. If we have no operations, what's the purpouse of the dataview?

packages/data/src/lib/data-view/data-view.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    });
+
+    it('should return an object with data aggregations', async () => {
+      const layer = new Layer('fake_source');
+      spyOn(layer, 'getViewportFeatures').and.returnValue(
+        Promise.resolve(sourceData)
```
VictorVelarde: Nice & clear. Do you think that we should group these 'layer'- tests under a big 'describe'? We will eventually have a 'source-' block, with other spies and mocks...

packages/data/src/lib/data-view/data-view.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  describe('groupBy', () => {
+    it('should return average of grouped values by column', async () => {
+      const layer = new Layer('fake_source');
+      spyOn(layer, 'getViewportFeatures').and.returnValue(
+        Promise.resolve(sourceDataToGroup)
+      );
+
+      const dataView = new DataView(layer, 'popEst');
+
+      expect(await dataView.groupBy('country')).toEqual([
```
VictorVelarde: I find a bit arbitrary to use AVG by default, without been explicit. In some variables SUM could also be a perfectly good default... In any case, I think the operation param should be mandatory, to avoid misunderstandings. Don't you think?

packages/data/webpack.config.js
```diff
@@ -0,0 +1,17 @@
...
+// Path and filename of your result bundle.
+// Webpack will bundle all JavaScript into this file
+const webpackBaseConfig = require('../../bundle/webpack.base');
+
+const webpackConfig = {
+  ...webpackBaseConfig,
+  output: {
+    path: path.resolve(__dirname, 'dist/umd'),
+    filename: 'index.min.js',
+    library: 'cartoViz',
```
VictorVelarde, line 12: `cartoData` instead?

packages/viz/src/lib/Layer.ts
```diff
@@ -1,20 +1,16 @@
...
-import {
-  ViewportFeaturesGenerator,
-  ViewportFeaturesOptions
-} from './interactivity/viewport-features/ViewportFeaturesGenerator';
+import { ViewportFeaturesGenerator } from './interactivity/viewport-features/ViewportFeaturesGenerator';
 import { CartoLayerError, layerErrorTypes } from './errors/layer-error';
 
-
-export class Layer implements StyledLayer {
+export class Layer extends WithEvents implements StyledLayer {
```
VictorVelarde: Nice!

packages/viz/src/lib/Layer.ts
```diff
@@ -144,19 +143,35 @@ export class Layer implements StyledLayer {
...
+    const events = {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      onViewportLoad: (...args: any) => {
+        // TODO(jbotella): Change typings
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const styleProperties = styleProps as any;
+
+        if (styleProperties.onViewportLoad) {
+          styleProperties.onViewportLoad(...args);
+        }
```
VictorVelarde: What's the purpose of this 3 lines?

packages/viz/src/lib/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -2,7 +2,6 @@ import { Deck, Viewport } from '@deck.gl/core';
 import { MVTLayer } from '@deck.gl/geo-layers';
 import { Matrix4 } from '@math.gl/core';
 import { selectPropertiesFrom } from '../../utils/object';
-import { AggregationTypes, applyAggregations } from './aggregations';
```
VictorVelarde, line 5: Nice, it's cleaner after splitting responsabilities with Dataviews...

VictorVelarde (commented):
The code and the feature looks nice.

I made some small comments / suggestions, that you might want to consider.

But the most relevant discussion is IMO the one related to parameters passed to `constructor`/ `groupBy`/ `aggregate`. I see your choice as obviously more flexible, but I would like to discuss a bit more on convenience / simplicity of that solution. If we pass all parameters to constructor it would be less flexible, but easier to use (and we could even remove the `groupBy`, having just a `calculate` method.

packages/data/src/lib/data-view/data-view.spec.ts
```diff
@@ -0,0 +1,173 @@
+import { CartoError } from '@carto/toolkit-core';
+import { Layer } from '@carto/toolkit-viz';
+import { DataView } from './data-view';
+import { AggregationTypes } from '../aggregations/aggregations';
+
+describe('DataView', () => {
```
VictorVelarde: We should probably create a tests folder like in other packages...

examples/dataviews/widget.html
```diff
@@ -0,0 +1,65 @@
...
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>GeoJSON Layer with CARTO Tiles Example</title>
+
+  <!-- Custom CSS -->
+  <link rel="stylesheet" type="text/css" href="https://api.tiles.mapbox.com/mapbox-gl-js/v1.0.0/mapbox-gl.css">
+  <link rel="stylesheet" href="https://libs.cartocdn.com/airship-style/v2.1.1/airship.css">
```
VictorVelarde: You can update to the latest changes (versions and panel style) from 'template', once that 8.1.7 update from Manu is merged on #124

packages/data/src/lib/data-view/data-view.spec.ts
jesusbotella: No, I think we should leave tests right next to their source files to make them easier to find.

I didn't like the split folder for tests, and tried to change it in the visualization module but I couldn't and now tried again and succeed. So I think we should do the same with all modules.

VictorVelarde: 🆗 I also like the idea of having them side by side, but then let's add a small task to make the same in other packages

simon-contreras-deel: But we already have in a separate folder... so we will have 2 ways. I would add tests in the __test__ folder by now and then, if we want to change them, do it in a different PR

packages/data/src/lib/operations/aggregation/aggregation.ts
```diff
@@ -0,0 +1,68 @@
...
+  [AggregationType.SUM](values: number[]) {
+    return values.reduce((sum, value) => sum + value, 0);
+  },
+
+  [AggregationType.PERCENTILE](values: number[], aggregationData: string[]) {
+    const percentile = parseInt(aggregationData[0], 10);
+
+    if (!Number.isInteger(percentile) && percentile > 0 && percentile < 100) {
+      // eslint-disable-next-line no-console
+      console.warn(
```
simon-contreras-deel: shouldn't we throw here?

packages/core/src/mixins/with-events.ts
```diff
@@ -0,0 +1,49 @@
+import mitt from 'mitt';
+import { CartoError } from '../error/carto-error';
+
+export abstract class WithEvents {
```
simon-contreras-deel, line 4: The name sounds a bit weird for me (I don't expect an event emitter), I would prefer using `Emitter` like `CARTOEmitter` or even `EventEmitter` like the node one

simon-contreras-deel (approved):
Just 2 comments at your discretion. More important IMO previous comment about tests folder

packages/data/src/lib/operations/aggregation/aggregation.ts
VictorVelarde: I do think so

packages/data/src/lib/aggregations/aggregations.ts
```diff
@@ -7,54 +7,43 @@ export enum AggregationTypes {
...
-      console.warn(
-        `[ViewportFeatures] ${aggregation} aggregation type not implemented`
-      );
-    }
-
-    result[aggregation] = aggregationFunction(values, aggregationData);
-  });
+  if (!aggregationFunction) {
+    // eslint-disable-next-line no-console
+    console.warn(`[Dataview] ${aggregation} aggregation type not implemented`);
```
VictorVelarde: Isn't it better to throw?

packages/data/src/lib/dataview/category/category.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        operation: AggregationType.AVG,
+        max: 50,
+        min: 15,
+        nullCount: 0
+      });
+    });
+  });
+});
+
+const sourceDataToGroup = [
```
VictorVelarde: picky: if it's only used in one test like in this case, I think it's more clear it's inside the test itself

packages/data/src/lib/dataview/dataview.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+      const dataUpdateSpy = jest.fn();
+      dataView.on('dataUpdate', dataUpdateSpy);
+
+      layer.emit('viewportLoad');
+
+      expect(dataUpdateSpy).toHaveBeenCalled();
+    });
+  });
+
+  // describe('aggregate', () => {
```
VictorVelarde: let's remove these comments, right?

packages/data/src/lib/dataview/dataview.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    this.bindEvents();
+  }
+
+  private bindEvents() {
+    this.registerAvailableEvents(['dataUpdate']);
+
+    if (this.dataSource instanceof Layer) {
+      this.dataSource.on('viewportLoad', () => {
+        this.onDataUpdate();
```
VictorVelarde, line 23: 👌

examples/dataviews/widget.html
```diff
@@ -51,10 +51,15 @@
...
+        { operation: 'count', operationColumn: 'pop_est' }
+      );
+
       dataview.on('dataUpdate', async () => {
-        const placesByContinent = await dataview.groupBy('continent', { operation: 'count' });
-        widgetElement.categories = placesByContinent;
+        const data = await dataview.getData();
+        widgetElement.categories = data.categories;
       });
     }
```
VictorVelarde: I'm gonna move this example to data, and update its libs references

VictorVelarde (changes requested):
LGTM. We have just 2 console.warn which I think we should turn into exceptions (and some minor comments to consider).

After that, feel free to merge

packages/core/src/mixins/with-events.ts
VictorVelarde: Not strong opinion on this, but I would leave it as it (eg Mapbox has Evented, which is pretty similar concept)

jesusbotella: Yeah, the thing is that it is not an event emitter per se, but a class that implements a mechanism to handle and manage events.

The initial approach with multiple inheritance was to extend the class like this:
```js
WithEvents(Dataview)
```

But that approach was not needed because all the classes that we needed to extend with this one didn't inherit from another class. I decided to leave the name though as it didn't look bad with:

```js
export class DataView extends WithEvents {}
```

packages/data/src/lib/operations/aggregation/aggregation.ts
jesusbotella: Yeah, probably, this remains from the past implementation where we could get multiple aggregations in the same method execution.

jesusbotella:
Hey @manmorjim! I think that any of your merges have removed some of my commits. I'm missing some tests and some code that I removed in:

https://github.com/CartoDB/toolkit/pull/118/commits/0e93e7074f88568de3c81a365ff1b848c0f9bdfa
https://github.com/CartoDB/toolkit/pull/118/commits/44df0d5689e28516f1f7845a2cafa35791d61310

packages/data/src/lib/dataview/dataview.spec.ts
jesusbotella: I think that you reviewed an outdated version of the pull request 🤔 . These comments have been removed since yesterday

VictorVelarde: Oh, sorry, My mistake. I was reviewing commit by commit to get a better understanding of some parts, and I thought I have switched to final changes...

VictorVelarde (approved):
LGTM. Let's merge it once we solve the interactivity conflict with @manmorjim

manmorjim:
This is ready to be merged! :rocket: 

git_merge

PR #119, opened 2020-05-22
[Popup & Interactivity] Default highlight styling

manmorjim:
Implemented default highlight styling:

default_highlight

### DIFF
--- a/packages/viz/src/lib/layer/LayerInteractivity.ts
+++ b/packages/viz/src/lib/layer/LayerInteractivity.ts
@@ -1,6 +1,6 @@
-import { Deck } from '@deck.gl/core';
+import { Deck, RGBAColor } from '@deck.gl/core';
 import { Popup, PopupElement } from '../popups/Popup';
-import { Style } from '../style/Style';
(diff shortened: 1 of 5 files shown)


packages/viz/src/lib/layer/LayerInteractivity.ts
```diff
@@ -180,13 +181,19 @@ export class LayerInteractivity {
 
     let clickStyleProps = {};
 
-    if (this._clickStyle) {
+    if (this._clickStyle === 'default') {
+      const defaultHighlightStyle = this._getDefaultHighlightStyle();
```
VictorVelarde, line 185: 🆗 I think we can keep both styles the same now, both in yellow for 'hovered' and 'clicked' features. We might decide to change that in the future... thoughts @borja-munoz ?

VictorVelarde (approved):
LGTM. Nice feature!

packages/viz/src/lib/layer/LayerInteractivity.ts
borja-munoz: Agree. Let's keep it like this for now.

PR #120, opened 2020-05-25
[Popup & Interactivity] Disabling cursor if no click nor hover interactions are set

manmorjim:
Added example `enable-disable-onhover-onclick.html ` in debug folder.


VictorVelarde (approved):
LGTM. Nice work!

PR #121, opened 2020-05-25
[Popup & Interactivity] Add D3 format

manmorjim:
- Update onhover example using D3 format

### DIFF
--- a/examples/interactivity/add-popups-onhover.html
+++ b/examples/interactivity/add-popups-onhover.html
@@ -85,7 +85,7 @@
         );
         await populationLayer.setPopupHover([
           'cartodb_id',
-          { attr: 'pop_2015', title: 'Population' }
+          { attr: 'pop_2015', title: 'Population', format: '~s' }
         ]);
         await populationLayer.addTo(deckMap);
       }
--- a/examples/interactivity/popups-format.html
+++ b/examples/interactivity/popups-format.html
@@ -106,7 +106,7 @@
             format: value => value.toString().replace(/000$/, 'K habitants')
           }
         ]);
-        await populationLayer.addTo(deckMap);
+        populationLayer.addTo(deckMap);
       }
       initialize();
     </script>
--- a/examples/interactivity/popups-format.html
+++ b/examples/interactivity/popups-format.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Add popup on hover example</title>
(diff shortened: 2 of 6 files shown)


examples/interactivity/add-popups-onhover.html
```diff
@@ -85,7 +85,7 @@ <h1 class="as-title">Pop-ups on hover</h1>
         );
         await populationLayer.setPopupHover([
           'cartodb_id',
-          { attr: 'pop_2015', title: 'Population' }
+          { attr: 'pop_2015', title: 'Population', format: '~s' }
```
VictorVelarde: Maybe we can add a comment to describe this is a d3.format. And a link to https://github.com/d3/d3-format for examples (as we will do in the docs in the future)

VictorVelarde (approved):
🚀

examples/interactivity/add-popups-onhover.html
VictorVelarde: Totally unrelated, but thinking of it, we could use title: null to completely hide the field, just displaying the 'attr' in the popup.

VictorVelarde:
As a feature to consider in the future. We're not giving datetime a lot of importance now, but we might need to generalize a bit the approach in the future, perphaps adding also `https://github.com/d3/d3-time-format`

/cc @borja-munoz for awareness

examples/interactivity/popups-format.html
```diff
@@ -0,0 +1,114 @@
...
+        });
+
+        const sizeBinStyle = new carto.viz.sizeBinsStyle('pop_2015', {
+          sizeRange: [1, 50]
+        });
+        const populationLayer = new carto.viz.Layer(
+          'world_population_2015',
+          sizeBinStyle
+        );
+        await populationLayer.setPopupHover([
```
VictorVelarde: Do we really need to use **await** here? The idea would be: avoid it if we can, and add it just if we really need to wait (for example, to ensure a proper order of layers when adding several to the map)

packages/viz/src/lib/popups/Popup.ts
```diff
@@ -228,21 +229,38 @@ function generatePopupContent(
...
-            elementValue = format.call(element, elementValue);
+            elementValue = format(elementValue);
+          } else if (format && typeof format === 'string') {
+            let formatter;
+
+            try {
+              formatter = d3Format(format);
+            } catch (err) {
+              throw new CartoPopupError(
+                `The format '${format}' is not a recognized D3 format`,
```
VictorVelarde, line 255: Nice!

VictorVelarde:
LGTM

PR #122, opened 2020-05-26
Add example of interactivity with multiple layers

manmorjim:


VictorVelarde (approved):
Looks good to me. As it is debug example, go for it.

If we ever need it to be official, we could maybe move all the buttons inside the right panel, for similarity with other examples

PR #123, opened 2020-05-26
Adapting popups & interactivity to the new event mechanism

manmorjim:


manmorjim:
I've just added small changes in the `Layer` class so only this class shoud be reviewed.

I also added the `mitt` dependency to the viz `package.json` because it needs it. I'm not sure about this, maybe a better option would be just exporting the `mitt.Handler` type?

examples/interactivity/popups-format.html
```diff
@@ -106,7 +106,7 @@ <h1 class="as-title">Pop-ups on hover with format</h1>
             format: value => value.toString().replace(/000$/, 'K habitants')
           }
         ]);
-        await populationLayer.addTo(deckMap);
+        populationLayer.addTo(deckMap);
```
VictorVelarde, line 109: removing what's not required will be great for the final user, thx for taking the time and effort!

jesusbotella:
I've just taken a look at this, and the synchronization that I built between the Dataviews and the Layer would be OK. So no more changes needed.

My point was to use that event management that I built to synchronize Deck.gl events and our events for popups and avoid having two different event management approaches (the more general one and the one for the interactivity). That way we can remove some repeated code and make it more manageable, but we can leave it for a refactor later as it would require a big effort.

manmorjim:
Changed Popup in order to allow register severals `onViewStateChange` events: https://github.com/CartoDB/toolkit/pull/123/commits/3fe53bc2e03ab030a45f2b428b32e3a270ffb231

packages/viz/src/lib/layer/LayerInteractivity.ts
```diff
@@ -21,122 +18,94 @@ export class LayerInteractivity {
...
 
   constructor(
     layer: StyledLayer,
     layerGetStyleFn: () => Style,
     layerSetStyleFn: (style: Style) => Promise<void>,
+    layerEmitFn: (type: string, event?: unknown) => void,
+    layerOnFn: EventHandler,
+    layerOffFn: EventHandler,
     hoverStyle?: Style | string,
     clickStyle?: Style | string
```
VictorVelarde: Picky comment: I feel like we could refactor this a bit into an options object, cause the number of params is now a bit high and some of them are optional

packages/viz/src/lib/layer/LayerInteractivity.ts
```diff
@@ -21,122 +18,94 @@ export class LayerInteractivity {
...
-          this._setCursorOnHover = true;
-        } else {
-          this._props.onHover = undefined;
-        }
-      }
-    } else {
-      const layerHandlerFn = (info: any, event: HammerInput) => {
-        const features = [];
-        const { coordinate, object } = info;
+  public fireOnEvent(eventType: EventType, info: any, event: HammerInput) {
```
VictorVelarde: Should we use a more specific naming for this enum, like `InteractivityEventType`?

manmorjim:
Refactor done!

VictorVelarde:
@manmorjim looks good to me. Let's ping @jesusbotella to check if we can now go and merge this into dataviews...

packages/viz/src/lib/popups/Popup.ts
```diff
@@ -55,6 +55,8 @@ export class Popup {
       this._parentElement = canvasElem.parentElement;
     }
 
+    // this._deckInstance.on('onViewStateChange', this._render.bind(this));
```
VictorVelarde: remove this comment?

jesusbotella (approved):
LGTM!

The only thing that I see is that you can call `off` method but it doesn't mean that you are unregistering the handler because it can be another handler that doesn't match the current ones. So not sure that we should always decrement the variable.

manmorjim:
> The only thing that I see is that you can call `off` method but it doesn't mean that you are unregistering the handler because it can be another handler that doesn't match the current ones. So not sure that we should always decrement the variable.

Yeah, good catch :+1: We should think about a mechanism to get all the handlers attached to an instance

PR #124, opened 2020-05-27
Upgrade deckgl package to 8.1.7

manmorjim:


VictorVelarde:
Looks good to me. Let me add a small change here on 'template.html' I was working on, for better airship use. The change, a bit more verbose, ensures the panel is still visible on the bottom when reducing the screen width, avoiding the 'default' responsive mode from airship, which assumes the use of tabs.

VictorVelarde (changes requested):
@manmorjim almost ready to go. Please, just make a quick check that the references on examples fit the ones from the 'template'. We have some  wrong & mixed ones, like mapbox 1.0 instead of mapbox 1.10.0.

IMO we can use this kind of 'global' examples updates to check there are no errors or differences among them.

In the future we'll also need to be homogeneous on what **carto accounts** are used. I would propose keeping them as they are now, but moving to just `public` in the short future (no cartoframes, no cartovl user...), exporting the datasets if required. Maybe we could also get more feedback regarding to examples / docs in further conversations with @borja-munoz

manmorjim:
> @manmorjim almost ready to go. Please, just make a quick check that the references on examples fit the ones from the 'template'. We have some wrong & mixed ones, like mapbox 1.0 instead of mapbox 1.10.0.

done!

VictorVelarde (approved):
Go for it! 😄

PR #126, opened 2020-05-28
Keep layer ordering

VictorVelarde:
We're always pushing the layer to the last position, every time we update it. By default, we should keep its current position. 

This PR just ensures that.

~I have added also a small refactor, affecting quite a lot of files in 1 commit: rename getLayerProps to getProps in Style class, which I think it's nicer.~ (reverted)

Note: 
I have made 'replaceDeckGLLayer' method public for the Layer. I think it makes sense to have it, but I'm open to suggestions.

@manmorjim I'm not sure how if we need to add any other change, in the context of popups management.

### DIFF
--- a/packages/viz/__tests__/Layer.spec.ts
+++ b/packages/viz/__tests__/Layer.spec.ts
@@ -46,25 +46,46 @@ describe('Layer', () => {
   });
 
   describe('Deck.gl integration', () => {
+    let deckInstanceMock: Deck;
+
+    beforeEach(() => {
+      const deck = {
+        props: {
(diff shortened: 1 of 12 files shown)


VictorVelarde:
We can wait for #123 and #118 to be merged, in order to ease merging.

VictorVelarde:
I'm gonna revert the refactor, so this is really small and we can merge it without colliding with any other PR WIP. Would you please @oleurud ?

simon-contreras-deel (approved):
:ok_hand:

PR #128, opened 2020-06-01
Use pointRadiusUnits prop & update to deck.gl@8.2.0-alpha.2

simon-contreras-deel:
Prepare style helpers to work with pointRadiusUnits property.

Do not merge until https://github.com/visgl/deck.gl/pull/4624 will be merged and released


examples/template.html
```diff
@@ -59,7 +59,7 @@ <h1 class="as-subheader">Basic example template</h1>
   </div>
 
   <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.10.0/mapbox-gl.js"></script>
-  <script src="https://unpkg.com/deck.gl@8.1.7/dist.min.js"></script>
+  <script src="https://unpkg.com/deck.gl@8.2.0-alpha.2/dist.min.js"></script>
```
VictorVelarde, line 62: Great, please also add the reference to the description in the body: <li class="as-list__item">DeckGL 8.1.7</li>

VictorVelarde:
Looks good regarding to pixel units, ~but we might have some unrelated issues with alpha.2 in other areas: I saw that `viewportFeatures` fails sistematically on https://toolkit-git-use-pointradiusunits-prop.carto-frontend.now.sh/examples/data/viewport-features.html, and we have to check that~

Forget about viewportFeatures, that example was not adapted to the new event system. I will fix it in another branch, so no worries.

VictorVelarde (approved):
💥

PR #129, opened 2020-06-02
Fix alpha in hex colors

simon-contreras-deel:


VictorVelarde (approved):
LGTM. Do we have any example (maybe in debug) using alpha, to check these values visually?.

Nice tests, by the way 👍

simon-contreras-deel:
You can check it here https://toolkit-git-fix-color-helpers-line-color.carto-frontend.now.sh/examples/_debug/styles/basic-style.html (example from next PR with this one already merged)

PR #130, opened 2020-06-02
[Popup & interactivity] Popup position issues

manmorjim:
This PR just fixes the issue related with the popup position when user changes zoom.


packages/viz/src/lib/popups/Popup.ts
```diff
@@ -55,15 +55,13 @@ export class Popup {
       this._parentElement = canvasElem.parentElement;
     }
 
-    // this._deckInstance.on('onViewStateChange', this._render.bind(this));
-    const { onViewStateChange } = this._deckInstance.props;
-
+    const { onAfterRender } = this._deckInstance.props;
     this._deckInstance.setProps({
-      onViewStateChange: (...args: any) => {
+      onAfterRender: (...args: any) => {
```
VictorVelarde, line 60: It looks like we force a new render... How did you come into this solution? Any tip or comment from vis.gl repo maybe?

manmorjim: Not really, I just found that event callback in the [doc of deckgl](https://deck.gl/#/documentation/deckgl-api-reference/deck) which says:

_"Called right after the canvas rerenders."_

By this way we have the last view and we can use it to unproject the coordinates to the new pixels location.

VictorVelarde (approved):
🚀

PR #131, opened 2020-06-02
Fix stroke color in color helpers for points and polygons

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/76418/styles-missing-strokecolor-in-colorbinsstyle-example

This PR fixes stoke lines in polygons and point when using color helpers

### DIFF
--- a/examples/_debug/styles/basic-style.html
+++ b/examples/_debug/styles/basic-style.html
@@ -93,10 +93,10 @@
         });
 
         const basicStyle = carto.viz.basicStyle({
-          // color: '#0000FF',
-          // size: 14,
-          // strokeColor: '#00FF00',
-          // strokeWidth: 10
+          color: '#0000FF',
+          size: 10,
+          strokeColor: '#0000FF44',
+          strokeWidth: 10
         });
         const layer = new carto.viz.Layer('sf_nbhd_crime', basicStyle);
         layer.addTo(deckMap);
--- a/examples/_debug/styles/color-categories-style.html
+++ b/examples/_debug/styles/color-categories-style.html
@@ -64,7 +64,7 @@
           }
         });
 
-        const style = carto.viz.colorCategoriesStyle('length_km');
+        const style = carto.viz.colorCategoriesStyle('type');
         const layer = new carto.viz.Layer('roads', style);
         layer.addTo(deckMap);
       }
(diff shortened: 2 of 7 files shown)


examples/_debug/styles/basic-style.html
```diff
@@ -93,10 +93,10 @@
         });
 
         const basicStyle = carto.viz.basicStyle({
-          // color: '#0000FF',
-          // size: 14,
-          // strokeColor: '#00FF00',
-          // strokeWidth: 10
+          color: '#0000FF',
+          size: 10,
+          strokeColor: '#0000FF44',
```
VictorVelarde, line 98: it works nicely!

examples/_debug/styles/color-categories-style.html
```diff
@@ -35,24 +35,41 @@
...
+            zoom: 11
+          }
+        });
+
+        const style = carto.viz.colorCategoriesStyle('category');
+        const layer = new carto.viz.Layer('sf_nbhd_crime', style);
+        layer.addTo(deckMap);
+      }
+
       function initializeLines() {
```
VictorVelarde, line 55: This produces an error if you activate it:
```
Uncaught (in promise) CartoStylingError: [Error] Current dataset has not categories for 'length_km'

```

Please let's double check, in case it's an underlying problem

packages/viz/src/lib/style/helpers/color-bins-style.ts
```diff
@@ -119,14 +120,27 @@ function calculateWithBreaks(
     return hexToRgb(colors[featureValueIndex]);
   };
 
+  let geomStyles;
+
+  if (geometryType === 'Line') {
```
VictorVelarde, line 125: 👍

examples/_debug/styles/color-categories-style.html
VictorVelarde: I guess this could be related to the use of a numerical field

VictorVelarde (approved):
LGTM. I just made a comment about a broken example in debug

examples/_debug/styles/color-categories-style.html
simon-contreras-deel: Yes, I am going to fix it

PR #132, opened 2020-06-03
Fix size continuous size range

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/76420/styles-sizecategories-and-sizecontinuous-still-not-working-100
Nothing to fix but the example. Also, opacity prop reviewed in helpers


VictorVelarde (approved):
LGTM. 🚀  !!!

PR #133, opened 2020-06-03
Category Widget

jesusbotella:
WIP.


packages/viz/src/lib/widgets/widget.ts
```diff
@@ -0,0 +1,55 @@
+import { CategoryDataView } from '@carto/toolkit-data';
+import { CartoError } from '@carto/toolkit-core';
+
+export class Widget {
+  protected element: HTMLElement;
+  protected dataView: CategoryDataView;
```
VictorVelarde, line 7: Should we use the baseclass here, just DataView?

packages/viz/src/lib/widgets/widget.ts
```diff
@@ -0,0 +1,55 @@
...
+  dataView: CategoryDataView
+) {
+  if (!element) {
+    throw new CartoError({
+      type: '[Widget]',
+      message: 'Element passed to Category Widget is not valid'
+    });
+  }
+
+  if (!(dataView instanceof CategoryDataView)) {
```
VictorVelarde, line 41: Oh, sorry, I see now it's WIP, specific for CategoryWidget...

PR #134, opened 2020-06-03
Formula dataview

VictorVelarde:
Add a simple FormulaDataView

There is a small example here: https://toolkit-git-feat-formula-dataview.carto-frontend.now.sh/examples/data/formula-data-view.html, with just some html as we don't effectively have a FormulaWidget (as we do with CategoryDataView --> CategoryWidget)


packages/data/src/lib/dataview/formula/formula.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    super(dataSource, column);
+
+    this.operation = operation;
+  }
+
+  async getData() {
+    const sourceData = await this.getSourceData([this.column]);
+
+    const filteredValues = sourceData.reduce(
```
jesusbotella: Hmm, here's one interesting thing. Should we cast strings into numbers? I did that in the CategoryDataView, but not sure if we should or not.

In fact we could make this simpler:

```js
const filteredValues = sourceData
  .map(dataObject => castToNumberOrUndefined(dataObject[this.column]))
  .filter(Boolean)
```

jesusbotella (commented):
It looks good! I think we can simplify the current `getData` function a bit, but apart from that looks good!

simon-contreras-deel (approved):
:ok_hand:

packages/data/src/lib/dataview/formula/formula.ts
VictorVelarde: Good point...

I would say that we should avoid casting strings, it could be risky, don't you think?. 

I'm gonna try to throw an Error if something not numerical is found in the top features

VictorVelarde: @jesusbotella I've added a small check on the first 10 features, raising an error if finding something different from null or number... After that, assuming that column already has numbers...

VictorVelarde:
Thx for merging it!

PR #135, opened 2020-06-04
[Dataview] - Category (II) - Source mode

manmorjim:


examples/data/dataview-category-source-mode.html
```diff
@@ -0,0 +1,163 @@
...
+              <as-category-widget class="as-p--16" heading="Cities">
+              </as-category-widget>
+            </div>
+          </div>
+        </div>
+      </main>
+    </div>
+
+    <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.10.0/mapbox-gl.js"></script>
+    <script src="https://unpkg.com/deck.gl@8.1.7/dist.min.js"></script>
```
VictorVelarde: We have already merged the upgrade to deck.gl alpha. Please upgrade your branch / examples to it

packages/data/src/lib/dataview/SourceDataView.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    this._source = newSource;
+    this.emit('optionChanged');
+  }
+
+  public get column() {
+    return this._column;
+  }
+
+  public set column(newColumn: string) {
```
VictorVelarde, line 37: Wouldn't a new column potentially retrigger the maps instantiation?

packages/data/src/lib/dataview/category/category.spec.ts
```diff
@@ -23,11 +23,10 @@ describe('DataView', () => {
             operationColumn: 'fake_operation_column'
           })
       ).toThrow(
-        new CartoError({
-          type: '[DataView]',
-          message:
-            'Operation property not provided while creating dataview. Please check documentation.'
-        })
+        new CartoDataViewError(
```
VictorVelarde, line 26: Fine, let's use the new specific error 🆗

packages/data/src/lib/dataview/index.ts
```diff
@@ -1 +1,2 @@
 export { CategoryDataView } from './category/category';
+export { CategorySourceDataView } from './category/CategorySourceDataView';
```
VictorVelarde, line 2: Ok, when we mix them both together we should probably rename 'category/category' to 'category/Category', if we have inside a class

packages/maps/src/Client.ts
```diff
@@ -53,7 +53,62 @@ export class Maps {
     return this.instantiateMap(mapConfig);
   }
 
-  private async instantiateMap(mapConfig: unknown) {
+  public static generateMapConfigFromSource(source: string) {
+    const type = source.search(' ') > -1 ? 'sql' : 'dataset';
+
+    return {
+      [type]: source,
+      vectorExtent: 2048,
```
VictorVelarde: We should probably extract 2048 from this method and also equivalent ones on `instantiateMapFrom` into some shared consts

VictorVelarde (commented):
Overall, it looks good. I just made some comments. It would be nice to merge this soon as it is (without unifying into a common Dataview). Having that merged in the branch, we should then start with the refactor, using something like a strategy pattern, based on a constructor parameter (an enum maybe with 'viewport', 'source' options?

Pinging @jesusbotella for awareness

packages/data/src/lib/dataview/SourceDataView.ts
manmorjim: Yup, that's the idea. In this version we are not "caching" the map instantiation yet so it'll be instantiated always but when we implement the caching mechanism changing the column will invalid the cache and trigger the map instantiation again.

VictorVelarde: Ok, just to be clear, that map instantiation is one specific for the dataviews, unrelated to the one for the map itself right? If that's the case, how does t behave if we do a setSource in the Layer? I guess it would have to be retriggered too or should the user recreate the dataviews with the new Source?. 

Please think a bit about these variations, so we can have a more detailed list of tasks for the next PR

manmorjim: Yes, the map instantiation is one specific for the dataviews.

About if the user does a `setSource`, that's a good point. ATM the user would have to recreate the dataviews with the new source but I think the option you suggested is better. We likely should add an event within Layer which is triggered when the user changed the source in order to instantiate the maps of dataview again.

VictorVelarde: Or at least an error in the meantime...

VictorVelarde:
We should close this one right? (in favour of https://github.com/CartoDB/web-sdk/pull/14)

PR #137, opened 2020-06-04
Getting some improvements from #128

simon-contreras-deel:
Getting some improvements added into #128 
- recover deleted code ensuring max and min radius and stroked props
- default sizeRange in points for size categories helper


VictorVelarde (approved):
LGTM

PR #138, opened 2020-06-05
Prevent adding the column carotbd_id as aggregation column

manmorjim:
This fixes the error `column reference "cartodb_id" is ambiguous` thrown by Maps API when we add the column `cartodb_id` whithin the popups.

### DIFF
--- a/packages/viz/src/lib/sources/CARTOSource.ts
+++ b/packages/viz/src/lib/sources/CARTOSource.ts
@@ -149,7 +149,8 @@ export class CARTOSource extends Source {
 
     const dimensions: Record<string, { column: string }> = {};
     fields.forEach(field => {
-      if (field.aggregation) {
+      // prevent using cartodb_id as aggregation column
+      if (field.aggregation && field.column !== 'cartodb_id') {
         dimensions[field.column] = { column: field.column };
       }
     });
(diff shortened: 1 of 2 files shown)


manmorjim:
Please, take a look at the comment I wrote in [CH](https://app.clubhouse.io/cartoteam/story/79584/popup-error-with-cartodb-id-column#activity-80626) also.

VictorVelarde (approved):
👌

packages/viz/src/lib/sources/CARTOSource.ts
```diff
@@ -149,7 +149,8 @@ export class CARTOSource extends Source {
 
     const dimensions: Record<string, { column: string }> = {};
     fields.forEach(field => {
-      if (field.aggregation) {
+      // prevent using cartodb_id as aggregation column
+      if (field.aggregation && field.column !== 'cartodb_id') {
```
VictorVelarde: how about a DEFAULT_ID_PROPERTY const?

For similar topics in the future we can check maps instantiation params in VL, like in_:
https://github.com/CartoDB/carto-vl/blob/57a55888a67d8e76bca8a259afaa992722a95c55/src/client/windshaft.js

simon-contreras-deel (commented):
I would filter it in Layer class (methods `_addPopupFields` && `_getStyleField`). At least, my idea was to avoid adding more logic in Source, since we could have to add same logic in other source types

manmorjim:
> I would filter it in Layer class (methods `_addPopupFields` && `_getStyleField`). At least, my idea was to avoid adding more logic in Source, since we could have to add same logic in other source types

good point, done!

simon-contreras-deel (approved):
:ok_hand:

PR #139, opened 2020-06-06
Fix examples

VictorVelarde:
- Fix viewportFeatures example, adapting it to the new getViewportFeatures interface and event system


simon-contreras-deel (approved):
:ok_hand:

